PR #7, opened 2024-01-31
ci: add macos to features check job

Noah-Kennedy:


Noah-Kennedy:
CI is frozen here due to how our CI reqs are specified.

inikulin:
@Noah-Kennedy I'm probably missing something, but GH now requires exact names of checks as a merge condition. It wasn't this way iirc. I'll change that, but that's weird UX tbh

inikulin (approved):

PR #10, opened 2024-02-01
ci: cleanup actions config

Noah-Kennedy:
This change cleans up our actions config to be less brittle when we make changes to CI, and to also be faster.

It uses several prebuilt actions from trusted providers for installing and using different tools, as well as caching artifacts. These changes should speed up our CI runs quite a bit.

It also reduces brittleness by specifying a new job, `pr-status-checks`, which we can specify PRs to depend on so that adding new jobs or editing matrixes won't break CI and require admin intervention.

It also breaks out our lint job into a set of smaller jobs to improve CI parallelism.


Noah-Kennedy:
@inikulin this one will require manual intervention from an admin. After this, `pr-status-checks` will be the only CI job we need to depend on.

fisherdarling (approved):

PR #11, opened 2024-02-02
telemetry: make logger verbosity public

evanrittenhouse:
I'm using [`slog-stdlog`](https://docs.rs/slog-stdlog/latest/slog_stdlog/) to read Quiche logs from Foundations, but I want to be able to configure the level of logs they emit. To do so, `slog-stdlog` exposes an [`init_with_level`](https://docs.rs/slog-stdlog/latest/slog_stdlog/fn.init_with_level.html) method - we obviously then need to get the verbosity level for the current logger.

I didn't see anything that lets us do this, as `Context::current_log()` just gets us a `SharedLog` which doesn't expose verbosity/`LoggerSettings` - if there's an easier way, please feel free to decline this PR.

### DIFF
--- a/foundations/src/telemetry/log/mod.rs
+++ b/foundations/src/telemetry/log/mod.rs
@@ -43,6 +43,12 @@ pub fn set_verbosity(level: Level) -> Result<()> {
     Ok(())
 }
 
+/// Gets the current log's verbosity.
+pub fn verbosity() -> LogVerbosity {
+    let harness = LogHarness::get();
+    harness.settings.verbosity
+}
+
 /// Returns current log as a raw [slog] crate's `Logger` used by Foundations internally.
 ///
 /// Can be used to propagate the logging context to libraries that don't use Foundations'


Noah-Kennedy:
You're failing a clippy check, can you remove the `.clone()` since this type is `Copy`?

Just amend the old commit right now since we can't squash on the reviewers side atm.

foundations/src/telemetry/log/mod.rs
```diff
@@ -43,6 +43,12 @@ pub fn set_verbosity(level: Level) -> Result<()> {
     Ok(())
 }
 
+/// Gets the current log's verbosity.
+pub fn get_verbosity() -> LogVerbosity {
+    let harness = LogHarness::get();
+    harness.settings.verbosity.clone()
```
Noah-Kennedy, line 49: Clippy doesn't like the clone as the verbosity is `Copy`.

```suggestion
    harness.settings.verbosity
```

Note that if you accept my suggestion, you will need to do a rebase and squash so that our history remains nicely bisectable.

Noah-Kennedy (changes requested):

Noah-Kennedy (approved):
Looks good!

foundations/src/telemetry/log/mod.rs
```diff
@@ -43,6 +43,12 @@ pub fn set_verbosity(level: Level) -> Result<()> {
     Ok(())
 }
 
+/// Gets the current log's verbosity.
+pub fn get_verbosity() -> LogVerbosity {
```
inikulin, line 47: Let's omit `get_` to make it cannonical

inikulin (changes requested):

inikulin (approved):

PR #12, opened 2024-02-02
telemetry: add tokio runtime metrics

Noah-Kennedy:
This PR introduces a new unstable API for collecting tokio runtime metrics which is gated behind `--cfg tokio_unstable` and `--cfg foundations_unstable`.

This doesn't add all metrics. In particular, tokio's poll duration histogram is not present since some additions might need to be made to our histogram API.


foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+            runtimes: Vec::new(),
+        }
+    }
+
+    /// Add a runtime to the monitor, optionally with a name and/or id in case you are monitoring multiple runtimes.
+    ///
+    /// Runtimes should be uniquely identifiable by both label and id.
+    pub fn add_runtime(
+        &mut self,
+        runtime_name: Option<Arc<str>>,
```
inikulin, line 86: Should we accept runtime ref as an argument instead to simplify API?

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+
+use crate::telemetry::metrics::{metrics, Counter, Gauge};
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use tokio::runtime::{Handle, RuntimeMetrics};
+
+/// Monitors a set of runtimes, allowing metrics to be periodically collected for each tracked runtime.
+#[derive(Default)]
+pub struct RuntimeMonitor {
+    runtimes: Vec<RuntimeHandle>,
```
inikulin, line 70: It doesn't feel like we need to have multiple runtimes in a single monitor. It would have be useful if monitor was a global singleton, but it's not, so users can just create a monitor per runtime if they need to.

However, I personally would prefer a global singleton (metrics are global anyway) approach (MemoryProfiler could be an inspiration). Also, it probably would be nice to auto sampling built-in with given frequency and providing a function to enforce sampling manually

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+
+    /// Record a sample of runtime metrics for each tracked runtime.
+    pub fn record_sample(&mut self) {
+        self.runtimes
+            .iter_mut()
+            .for_each(RuntimeHandle::record_sample)
+    }
+}
+
+struct RuntimeHandle {
```
inikulin, line 119: Let's split monitor, this and metrics definitions into separate modules

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+//!
+//! Foundations provides a helper API for monitoring Tokio runtimes using the
+//! [`foundations metrics api`](crate::telemetry::metrics).
+//! This helper API allows users to track one or more runtimes using a [`RuntimeMonitor`] object and collect their
+//! metrics periodically using [`RuntimeMonitor::record_sample`].
+//!
+//! ## Metrics
+//! | Metric                                           | Source                                                              | Labels (? indicates optional)          |
+//! |:-------------------------------------------------|:--------------------------------------------------------------------|:---------------------------------------|
+//! | tokio_runtime_workers                            | [`tokio::runtime::RuntimeMetrics::num_workers`]                     | runtime_name?, runtime_id?             |
```
inikulin, line 11: Let's just make metrics structs public. Yes, users could modify them manually, but it's not a big deal. It will be easier to have regular rust docs for them are in sync with metric info than maintaining with markdown table

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
+//! Tokio Runtime Metrics
```
inikulin, line 1: Don't capitilise each word

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+//!
+//! Foundations provides a helper API for monitoring Tokio runtimes using the
```
inikulin, line 3: Add a section explaining the build configuration that it needs

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+    pub fn add_runtime(
+        &mut self,
+        runtime_name: Option<Arc<str>>,
+        runtime_id: Option<usize>,
+        handle: &Handle,
+    ) {
+        if self.runtimes.iter().any(|handle| {
+            (handle.runtime_name.is_some() != runtime_name.is_some())
+                || (handle.runtime_id.is_some() != runtime_id.is_some())
+        }) {
```
inikulin, line 93: Simplify this condition by splitting it in bool vars with meaningful names

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+            panic!("If you specify a runtime name or ID for one runtime, that setting must be specified for all");
+        }
+
+        if self.runtimes.iter().any(|handle| {
+            (handle.runtime_name.is_some() && handle.runtime_name == runtime_name)
+                || (handle.runtime_id.is_some() && handle.runtime_id == runtime_id)
```
inikulin, line 99: Same as above for the condition

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+        &mut self,
+        runtime_name: Option<Arc<str>>,
+        runtime_id: Option<usize>,
+        handle: &Handle,
+    ) {
+        if self.runtimes.iter().any(|handle| {
+            (handle.runtime_name.is_some() != runtime_name.is_some())
+                || (handle.runtime_id.is_some() != runtime_id.is_some())
+        }) {
+            panic!("If you specify a runtime name or ID for one runtime, that setting must be specified for all");
```
inikulin, line 94: Don't panic, return `BootstrapError` instead

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+        let metrics = self.handle.metrics();
+
+        self.record_global_metrics(&metrics);
+
+        for worker_idx in 0..metrics.num_workers() {
+            self.record_worker_metrics(&metrics, worker_idx);
+        }
+    }
+
+    fn record_global_metrics(&mut self, metrics: &RuntimeMetrics) {
```
inikulin, line 136: make fn name consistent with how metrics are called elsewhere. I think we can name them `general_runtime_metrics` or `core_runtime_metrics or smthg like

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+        runtime_id: Option<usize>,
+    ) -> Counter;
+
+    /// Counter of forced yields due to task budgeting.
+    pub(super) fn budget_forced_yields_total(
+        runtime_name: &Option<Arc<str>>,
+        runtime_id: Option<usize>,
+    ) -> Counter;
+
+    pub(super) fn io_driver_fd_registrations_total(
```
inikulin, line 277: All the metrics need docs

foundations/src/telemetry/tokio/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+//! ```
+
+use crate::telemetry::metrics::{metrics, Counter, Gauge};
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use tokio::runtime::{Handle, RuntimeMetrics};
+
+/// Monitors a set of runtimes, allowing metrics to be periodically collected for each tracked runtime.
+#[derive(Default)]
+pub struct RuntimeMonitor {
```
inikulin, line 69: Move the usage example here and link it in the module docs

foundations/Cargo.toml
```diff
@@ -77,6 +77,12 @@ telemetry = [
     "telemetry-server",
 ]
 
+tokio-metrics = [
```
inikulin, line 80: Let's rename it to `tokio-runtime-metrics`

inikulin: and the module as well

foundations/Cargo.toml
```diff
@@ -77,6 +77,12 @@ telemetry = [
...
+    "tokio/rt",
+    "metrics",
+    "serde/rc",
+]
```
inikulin, line 84: Should it be in any of the default feature sets? If not, we need to explain why not in docs

foundations/Cargo.toml
inikulin, line 80: Needs to be added in lib.rs docs describing features along with explanation of how to enable it (as it's different from the rest)

inikulin (changes requested):

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: I think a global singleton monitor approach is a good idea here.

We definitely _do_ need to support multiple runtimes for this API, especially if it is singleton.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: I really, really, really do not want to make them public as people will absolutely try to edit them manually.

Also, this markdown table is actually quite easy to maintain.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: No objections to this on my end.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: Thanks, this was an oversight, I meant to have docs for these

foundations/Cargo.toml
Noah-Kennedy: I don't think we should put it in there until we stabilize this. When we do, we can revisit this.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: Generally, it is better to have overall examples that show how a whole API can be used be located in the module docs because those will display much more prominently on docs.rs.

foundations/Cargo.toml
Noah-Kennedy: No objections here, I don't really care all that much what color we paint the bikeshed in this case.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: I'm going to do this at the top-level in lib.rs

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: I don't think this should be treated as a runtime error. If I could make this a compilation error without making the API painful, I would do so.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: Thanks, I really hate the current name.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: done, and left a link in here.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: done

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: I've relaxed the rules and simplified the conditionals.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: I've made this an assertion.

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: renamed to tokio_runtime_core

foundations/src/telemetry/tokio/mod.rs
Noah-Kennedy: I've pushed up a static version.

Now, let's bikeshed the new names!

foundations/src/telemetry/tokio_runtime_metrics/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+pub fn record_runtime_metrics_sample() {
+    MONITOR
+        .lock()
+        .runtimes
+        .iter_mut()
+        .for_each(RuntimeHandle::record_sample)
+}
+
+#[derive(Default)]
+struct RuntimeMonitor {
```
inikulin, line 109: Now when all API is exposed as functions that work with statics we can remove this struct and just store a static vector. At the same time `RuntimeHandle` is ambiguous name, so probably let's rename `RuntimeHandle` to `RuntimeMonitor`

foundations/src/telemetry/tokio_runtime_metrics/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+//!
+//! [`foundations metrics api`]: crate::telemetry::metrics
+
+use crate::telemetry::tokio_runtime_metrics::runtime_handle::RuntimeHandle;
+use parking_lot::Mutex;
+use std::sync::Arc;
+use tokio::runtime::Handle;
+
+mod metrics;
+mod runtime_handle;
```
inikulin, line 69: Nit: mods before imports

inikulin (changes requested):

foundations/src/telemetry/tokio_runtime_metrics/mod.rs
Noah-Kennedy: I left it this way since this seems slightly cleaner, but I don't really care too much either way here so I'll make this change.

foundations/src/telemetry/tokio_runtime_metrics/mod.rs
Noah-Kennedy: I did the other thing since I wasn't sure what side of the style war you were on, and the imports-first camp seems to have won AFAICT.

Good to know you're a fellow fifth-columnist.

Noah-Kennedy:
One thing I realized - we probably need a way to deregister runtimes or at least to drop them if the runtime goes away.

foundations/src/telemetry/tokio_runtime_metrics/mod.rs
inikulin: I mean we consistently use exactly this style in all our codebases as far as I can tell...

Noah-Kennedy: define "consistently"

Noah-Kennedy:
@inikulin I've added deregistration and addressed the last bit of feedback.

inikulin (approved):

PR #14, opened 2024-02-04
telemetry: change LogVerbosity to an enum, use it throughout public API

### LINKED ISSUE #13: For API consistency make `LogVerbosity` an enum and use it across the API

evanrittenhouse:
Closes #13.

### DIFF
--- a/foundations/src/telemetry/log/init.rs
+++ b/foundations/src/telemetry/log/init.rs
@@ -140,7 +140,7 @@ where
         drain,
         FieldRedactFilterFactory::new(settings.redact_keys.clone()),
     );
-    let drain = drain.filter_level(*settings.verbosity);
+    let drain = drain.filter_level(settings.verbosity.into());
 
     RateLimitingDrain::new(drain, settings)
 }
--- a/foundations/src/telemetry/settings/logging.rs
+++ b/foundations/src/telemetry/settings/logging.rs
@@ -122,30 +122,3 @@ pub struct LogVolumeMetricSettings {
     /// Whether to enable log volume metrics
     pub enabled: bool,
 }
-
-#[cfg(feature = "settings")]
-mod with_settings_feature {
-    use super::*;
-
-    impl<'de> Deserialize<'de> for LogVerbosity {
-        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-        where
-            D: Deserializer<'de>,
-        {
-            Level::from_str(&String::deserialize(deserializer)?)
-                .map_err(|_| de::Error::custom("incorrect verbosity level"))
-                .map(LogVerbosity)
-        }
-    }
-
-    impl Serialize for LogVerbosity {
-        fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
-        where
-            S: Serializer,
-        {
-            s.serialize_str(self.0.as_str())
-        }
-    }
-
-    impl Settings for LogVerbosity {}
-}
--- a/foundations/src/telemetry/settings/logging.rs
+++ b/foundations/src/telemetry/settings/logging.rs
@@ -1,13 +1,10 @@
 use crate::telemetry::settings::rate_limit::RateLimitingSettings;
 use crate::utils::feature_use;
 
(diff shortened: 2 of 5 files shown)


foundations/src/telemetry/log/init.rs
```diff
@@ -137,7 +137,7 @@ where
         drain,
         FieldRedactFilterFactory::new(settings.redact_keys.clone()),
     );
-    let drain = drain.filter_level(*settings.verbosity);
+    let drain = drain.filter_level((&settings.verbosity).into());
```
inikulin, line 140: We needed deref before as we were wrapping `slog::Level`. Considering that now we have `into` and verbosity is a `Copy` type, we can just have `settings.verbosity.into()` here

inikulin (changes requested):

foundations/src/telemetry/settings/logging.rs
```diff
@@ -66,21 +65,58 @@ pub enum LogFormat {
     Json,
 }
 
-/// Verbosity level of the log.
+/// Log verbosity levels which match 1:1 with [`slog::Level`].
 #[derive(Clone, Debug, Copy)]
-pub struct LogVerbosity(pub Level);
+pub enum LogVerbosity {
```
inikulin, line 70: This needs settings attrs as for other types above

foundations/src/telemetry/settings/logging.rs
```diff
@@ -66,21 +65,58 @@ pub enum LogFormat {
...
+    /// See [`slog::Level::Debug`].
+    Debug,
+    /// See [`slog::Level::Trace`].
+    Trace,
+}
 
 impl Default for LogVerbosity {
     fn default() -> Self {
-        Self(Level::Info)
+        Self::Info
```
inikulin, line 87: nit: you can use derive `Default` in combination with `#[default]` attribute for the variant instead

inikulin (changes requested):

foundations/src/telemetry/settings/logging.rs
evanrittenhouse: Hmm, isn't there an `impl Settings` already? https://github.com/cloudflare/foundations/pull/14/files#diff-9eae3d0e9cd889fe453ebfadef152fa2ab92eae3cbb086ac9e32c7dbd46b757bR158

inikulin: Right, was hidden by the diff for me. But now, when we have an enum and not a wrapper, we can do it via macro

evanrittenhouse: Got it, done

foundations/src/telemetry/settings/logging.rs
```diff
@@ -66,21 +63,52 @@ pub enum LogFormat {
...
-pub struct LogVerbosity(pub Level);
+/// Log verbosity levels which match 1:1 with [`slog::Level`].
+#[cfg_attr(
+    feature = "settings",
+    settings(crate_path = "crate", impl_default = false)
+)]
+#[cfg_attr(not(feature = "settings"), derive(Clone, Debug))]
+#[derive(Copy, Default)]
+pub enum LogVerbosity {
+    /// See [`slog::Level::Critical`].
```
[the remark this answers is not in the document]
evanrittenhouse: Unfortunately the `slog` docs aren't very descriptive: https://docs.rs/slog/latest/slog/enum.Level.html#variant.Critical.

Maybe we just say "Matches `slog::Level::Critical`?

foundations/src/telemetry/settings/logging.rs
```diff
@@ -66,21 +63,52 @@ pub enum LogFormat {
     Json,
 }
 
-/// Verbosity level of the log.
-#[derive(Clone, Debug, Copy)]
-pub struct LogVerbosity(pub Level);
+/// Log verbosity levels which match 1:1 with [`slog::Level`].
```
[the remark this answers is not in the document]
evanrittenhouse: We may want to remove this note entirely depending on what we decide to do with my comment above

PR #22, opened 2024-02-12
Implement RangeGauge & GaugeGuard types

cbranch:


cbranch:
Updated to use the `fetch_max` and `fetch_min` atomic ops instead of `compare_exchange`.

foundations/src/telemetry/metrics/gauge.rs
```diff
@@ -0,0 +1,170 @@
...
+impl TypedMetric for RangeGauge {
+    const TYPE: MetricType = MetricType::Gauge;
+}
+
+impl EncodeMetric for RangeGauge {
+    fn encode(&self, mut encoder: Encoder) -> Result<(), std::io::Error> {
+        let current = self.get();
+        // Getting the current values of the metric resets the min/max
+        let min = self.min.swap(current, Ordering::Relaxed);
+        let max = self.max.swap(current, Ordering::Relaxed);
```
inikulin, line 84: What always bothered me acquiring values of multiple atomics is that it's always data race-prone, so you end up having non-atomic values under RwLock. Though, metrics could be best effort in that regard. But can you think of any scenarios where things can get wrong when reported counters are out of sync with actual values?

foundations/src/telemetry/metrics/mod.rs
```diff
@@ -21,8 +21,11 @@ pub(super) mod init;
 #[doc(hidden)]
 pub mod internal;
 
+mod gauge;
```
inikulin, line 24: Nit: non-public mods before publics

foundations/src/telemetry/metrics/mod.rs
```diff
@@ -373,3 +376,49 @@ impl MetricConstructor<TimeHistogram> for HistogramBuilder {
         TimeHistogram::new(self.buckets.iter().cloned())
     }
 }
+
+/// Increments a gauge metric when created, and decrements it when dropped.
+pub struct GaugeGuard<G: GenericGauge>(G);
```
inikulin, line 381: Let's move to `gauge.rs`

inikulin (changes requested):

foundations/src/telemetry/metrics/gauge.rs
cbranch: The metric expects the obvious invariant that min<=current<=max.
Swapping the min/max with a potentially stale version of current could break that invariant, so the old version was not good enough.

I really want to avoid wrapping the whole metric in a mutex, in particular I don't want inc/dec locked. But locking read/reset is OK as there should only be one metric scrape at a time. The invariant only matters when you scrape it, so we can 'fix up' the values with atomic ops after reading them and attempting to reset.

cbranch:
@nox do you have some time to check this one over?

nox (approved):
This looks good to me.

inikulin (approved):

PR #34, opened 2024-03-11
Add support for OTLP tracing reporting

inikulin:
It's recommended to review the commits in the PR separately


foundations/src/telemetry/driver.rs
```diff
@@ -0,0 +1,112 @@
...
+///
+/// [security syscall-related]: `crate::security`
+pub struct TelemetryDriver {
+    #[cfg(feature = "telemetry-server")]
+    server_addr: Option<SocketAddr>,
+
+    #[cfg(feature = "telemetry-server")]
+    server_fut: Option<TelemetryServerFuture>,
+
+    tele_futures: FuturesUnordered<BoxFuture<'static, BootstrapResult<()>>>,
```
cbranch, line 31: if this is always going to have one future, why is this different from `server_fut`

inikulin: No, in the upcoming PRs it will be also used to drive exporters.

cbranch (approved):

foundations/src/telemetry/driver.rs
```diff
@@ -0,0 +1,112 @@
...
+        let tele_futures_poll = Pin::new(&mut self.tele_futures).poll_next(cx);
+
+        if let Poll::Ready(Some(res)) = tele_futures_poll {
+            ready_res.push(res);
+        }
+
+        if ready_res.is_empty() {
+            Poll::Pending
+        } else {
+            Poll::Ready(ready_res.into_iter().collect())
```
Noah-Kennedy, line 109: Is completing after the first polled future finishes the intended behavior here?

inikulin: Yes. There are 2 cases where either of those complete:
1) TelemetryServer received a graceful shutdown signal. Once it finishes we wrap up all operations, including telemetry reporting
2) Any of those error - one of the components of the telemetry system is not functioning anymore, we need to terminate.

Noah-Kennedy (approved):

PR #40, opened 2024-03-26
Improve metrics bind error message

joel0:
Previously:
```
app_main: Err: Address already in use (os error 98)
```
Now:
```
app_main: Err: binding to socket [::]:8000

Caused by:
    Address already in use (os error 98)
```

### DIFF
--- a/foundations/src/telemetry/server.rs
+++ b/foundations/src/telemetry/server.rs
@@ -2,7 +2,7 @@
 use super::metrics;
 use super::settings::TelemetrySettings;
 use crate::{BootstrapResult, Result};
-use anyhow::anyhow;
+use anyhow::{anyhow, Context as _};
 use futures_util::future::BoxFuture;
 use futures_util::ready;
 use futures_util::FutureExt;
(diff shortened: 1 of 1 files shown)


nmldiegues (approved):

foundations/src/telemetry/server.rs
```diff
@@ -2,7 +2,7 @@
 use super::metrics;
 use super::settings::TelemetrySettings;
 use crate::{BootstrapResult, Result};
-use anyhow::anyhow;
+use anyhow::{anyhow, Context as AnyhowContext};
```
inikulin, line 5: Nit: you don't really need an alias for traits that you just want to bring in scope: `Context as _`.

inikulin (changes requested):

foundations/src/telemetry/server.rs
joel0: Ah, that is good to know.

inikulin (approved):

PR #47, opened 2024-06-12
ZTC-1478: track the number of times a logger has been replaced with a child logger

bcottrell8738:
Every time we call the add_fields! macro, it adds one to the depth of the nested structure of Arcs inside the logger object. If the structure gets too deeply nested, it causes a stack overflow on drop.

Fixing this to make it keep the structure flat rather than nested would be difficult; so, this commit makes it keep track of how many times it has replaced a logger with a child logger, and it stops you (with a panic) before you reach a dangerous level of nesting.

The panic string is intended to be informative enough so that you get a hint as to where to go look in your code.

### DIFF
--- a/foundations/src/telemetry/log/internal.rs
+++ b/foundations/src/telemetry/log/internal.rs
@@ -1,11 +1,93 @@
 use super::init::LogHarness;
 use crate::telemetry::scope::Scope;
 use slog::{Logger, OwnedKV, SendSyncRefUnwindSafeKV};
+use std::ops::Deref;
+use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::Arc;
 
+pub const MAX_LOG_GENERATION: u32 = 1000;
+pub const EXCEEDED_MAX_LOG_GENERATION_ERROR: &str = "foundations::telemetry::log: maximum logger generation exceeded (are add_fields! or set_verbosity being called in a loop?)";
+
+// The slog_logger() function exposes Arc<RwLock<Logger>> as a public
+// interface, so we can't store the generation number where it would make the
+// most sense to store it, which would be inside the RwLock, right next to
+// the Logger.
+//
+// Doing it this way means we lose the enforced atomicity between updating the
+// Logger and updating the generation number. In general, we try to only read
+// or write the generation number while we hold the appropriate lock on the
+// Logger. It is not mission-critical that we do this. The generation number
+// is just a guard rail to prevent us from going wild. Even if there is some
+// race condition and we're off by one or two, it won't affect its ability to
+// do its job.
+//
 // NOTE: we intentionally use a lock without poisoning here to not
 // panic the threads if they just share telemetry with failed thread.
-pub(crate) type SharedLog = Arc<parking_lot::RwLock<Logger>>;
+pub(crate) type SharedLog = Arc<LogContext>;
+
+#[derive(Debug)]
+pub struct LogContext {
+    // The logger itself. This is the most important part of this struct.
+    pub logger: Arc<parking_lot::RwLock<Logger>>,
+
+    // Generation number, incremented every time the logger is replaced with
+    // a child of itself. Accuracy is not critical as this is only used as a
+    // safety check. Nevertheless, when possible, please be holding the
+    // logger's lock when you read or write the generation number.
+    pub(crate) generation: AtomicU32,
+}
+
+impl LogContext {
+    // Create a new LogContext based on a fresh logger. The generation number
+    // is initialized to zero.
+    pub(crate) fn new(logger: Logger) -> Self {
+        Self {
+            logger: Arc::new(parking_lot::RwLock::new(logger)),
+            generation: AtomicU32::new(0),
+        }
+    }
+
+    // Update this LogContext's logger object, and, atomically, increment
+    // the generation number. This mutates 'self' using interior mutability,
+    // even though it only takes a shared reference to self.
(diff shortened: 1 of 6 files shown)


foundations/src/telemetry/log/internal.rs
```diff
@@ -1,11 +1,67 @@
...
+
+// The slog_logger() function exposes Arc<RwLock<Logger>> as a public
+// interface, so we can't store the generation number where it would make the
+// most sense to store it, which would be inside the innermost RwLock, right
+// next to the Logger. So we need two nested levels of Arcs and RwLocks.
+//
 // NOTE: we intentionally use a lock without poisoning here to not
 // panic the threads if they just share telemetry with failed thread.
-pub(crate) type SharedLog = Arc<parking_lot::RwLock<Logger>>;
+pub(crate) type SharedLog = Arc<parking_lot::RwLock<LogContext>>;
```
bcottrell8738, line 17: The comment above explains why we cannot get away with just a single `Arc` and a single `RwLock`, with both the `Logger` and the `u32` living underneath the `RwLock`.

If we really wanted to, we could get rid of the outer `RwLock`, and just have `SharedLog` be defined as an `Arc<LogContext>`. That would mean we would need interior mutability inside `LogContext` for updating the generation number; so that could be changed to an `AtomicU32`.

I initially did it that way; but I found myself writing voluminous comments everywhere in order to justify/excuse what I was doing. It felt like a code smell, to me. So I added the extra `RwLock` layer. I feel like this way of doing it is more readily comprehensible to people who might be reading the code.

OmegaJak: Personally, I think it would be better to use an `AtomicU32` on `generation`. It's weird to me to have a struct that has only partial interior mutability, as it forces what you did here - two layers of locks. Since we're forced into the `Arc<RwLock<Logger>>` for storing the logger itself, I think the next best choice is making the other field(s) on `LogContext` have their own locks.

One benefit is that it would reduce work on the hot path of actually writing logs, while it would probably be a similar amount of work (not sure how locking a RwLock compares to incrementing + reading an atomic) on the `add_fields!` path.

foundations/src/telemetry/log/internal.rs
```diff
@@ -36,7 +91,6 @@ pub fn current_log() -> SharedLog {
 
 pub(crate) fn fork_log() -> SharedLog {
     let parent = current_log();
-    let log = parent.read().new(slog::o!());
-
-    Arc::new(parking_lot::RwLock::new(log))
+    let log_lock = parent.read();
```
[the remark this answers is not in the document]
bcottrell8738: I'm not sure why it wouldn't be functionally the same. Calling `Logger::new()` with an empty KV will clone the drain and add zero fields to the KV. Which is basically what `<LogContext as Clone>::clone` will do, too.

I'm open to being convinced otherwise! but I will leave this as is for now.

foundations/src/telemetry/log/internal.rs
OmegaJak: I think this is an improvement. As you note, it's not mission critical that generation is exactly in sync - we don't care if it triggers at 1005 and not 1000, really. Maybe we could even used `Ordering::Relaxed`, but I'm no expert on what exactly that does and whether it's safe. I really like that we don't have an extra lock on the hot path of logging now. `overwrite_from` is a little gross, but I think it's worth it.

foundations/src/telemetry/log/internal.rs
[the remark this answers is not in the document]
OmegaJak: Ah - I mistakenly thought `o!` was making a new output or some such, not the KV. Looking at slog's source now, I agree, I think you're right and what you're doing is functionally the same.

foundations/src/telemetry/log/internal.rs
bcottrell8738: OK! I'm glad we did this, rather than just me making assumptions about what would be best! :-)

jgpaiva (approved):
Nice, this looks way better than if I had done it.
My only comment is around documenting this new behaviour, maybe the documentation for `add_fields` and `set_verbosity` should warn the user about the potential panic lurking inside?

foundations/src/telemetry/log/internal.rs
nmldiegues: With the current implementation, the `generation` field is indeed accessed always under the logger lock, which creates the necessary `happens-before` relationship. Those atomic accesses could be replaced by relaxed operations without  any impact.

As for what those mean, basically it's how strong of a memory barrier the compiler needs to issue depending on the architecture of the processor. For e.g., with strong consistency you don't really get any more expensive than with relaxed in x86_64.  But for ARM, strong consistency is more expensive than relaxed because the memory model of the processor is weaker than x86_64.

As for atomic increments vs lock acquisition, the former may be cheaper in some architectures (e.g. x86_64 has fetch_and_inc which is commutative and much more scalable in terms of concurrency than compare_and_swap, with the latter being the primitive used for locks).

nmldiegues (approved):

bcottrell8738:
> Nice, this looks way better than if I had done it. My only comment is around documenting this new behaviour, maybe the documentation for `add_fields` and `set_verbosity` should warn the user about the potential panic lurking inside?

How does that look?

jgpaiva:
> > Nice, this looks way better than if I had done it. My only comment is around documenting this new behaviour, maybe the documentation for `add_fields` and `set_verbosity` should warn the user about the potential panic lurking inside?
> 
> How does that look?

LGTM!

PR #48, opened 2024-06-14
Release 3.4.0

bcottrell8738:
My motivation for proposing a release is that I would like to get https://github.com/cloudflare/foundations/pull/47

I was figuring a "minor" bump for that, but I'm open to different viewpoints.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "3.3.0"
+version = "4.0.0"
 repository = "https://github.com/cloudflare/foundations"
 edition = "2021"
 authors = ["Cloudflare"]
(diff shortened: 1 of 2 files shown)


nmldiegues:
Clippy isn't happy

fisherdarling (changes requested):
Can we run `cargo-semver-checks` to determine the level of breaking changes? Since we don't track manually what's breaking or not, it's hard to tell from the RELEASE_NOTES.

bcottrell8738:
> Can we run `cargo-semver-checks` to determine the level of breaking changes? Since we don't track manually what's breaking or not, it's hard to tell from the RELEASE_NOTES.

cargo semver-checks says it needs to be a major version (!) -- I don't know if that also means we want more coordination around the timing of this. I was just hoping to get my change in there, but maybe my change should take a back seat.

fisherdarling:
> > Can we run `cargo-semver-checks` to determine the level of breaking changes? Since we don't track manually what's breaking or not, it's hard to tell from the RELEASE_NOTES.
> 
> cargo semver-checks says it needs to be a major version (!) -- I don't know if that also means we want more coordination around the timing of this. I was just hoping to get my change in there, but maybe my change should take a back seat.

Like it's suggesting `4.0.0`? That's fine! Whatever the semver gods require.

PR #49, opened 2024-06-19
ZTC-1545: Error during settings deserialization if YAML contains unused keys

OmegaJak:
The goal is to help users of foundations catch possible configuration mistakes which, depending on the context, could have serious consequences. For example, if a config field is renamed, but the name is not changed in the deserialized YAML file, foundations will use the defaults for the renamed config value, which may differ substantially from the previous value. With this change, if the feature is enabled, foundations will return an error on deserializing the YAML file since it contains an unused key.

This takes a different approach from #43. This PR results in an error instead of a logged warning, making it much harder to ignore.

If the opt-in foundations feature `settings-deny-unknown-fields` is enabled, all settings containers using the `#[settings]` macro will have the `#[serde(deny_unknown_fields)]` attribute set on them.

### DIFF
--- a/foundations-macros/src/settings.rs
+++ b/foundations-macros/src/settings.rs
@@ -26,6 +26,8 @@ struct Options {
     impl_debug: bool,
     #[darling(default = "Options::default_crate_path")]
     crate_path: Path,
+    #[darling(default = "Options::default_deny_unknown_fields")]
+    deny_unknown_fields: bool,
 }
 
 impl Options {
@@ -40,6 +42,10 @@ impl Options {
     fn default_crate_path() -> Path {
         parse_quote!(::foundations)
     }
+
+    fn default_deny_unknown_fields() -> bool {
+        true
+    }
 }
 
 impl Default for Options {
@@ -48,6 +54,7 @@ impl Default for Options {
             impl_default: Options::default_impl_default(),
             impl_debug: Options::default_impl_debug(),
             crate_path: Options::default_crate_path(),
+            deny_unknown_fields: Options::default_deny_unknown_fields(),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


foundations-macros/src/settings.rs
```diff
@@ -883,6 +913,26 @@ mod tests {
...
         };
 
+        #[cfg(feature = "settings-deny-unknown-fields")]
+        let expected = add_deny_unknown_fields_attr(expected);
+
         assert_eq!(actual, expected);
     }
+
+    #[cfg(feature = "settings-deny-unknown-fields")]
+    fn add_deny_unknown_fields_attr(mut code: String) -> String {
```
inikulin, line 923: I would prefer it to be done via the macro rather than code string substitution

foundations/Cargo.toml
```diff
@@ -55,6 +55,9 @@ settings = [
     "dep:indexmap",
 ]
 
+# Adds #[serde(deny_unknown_fields)] to all settings containers using #[settings]
+settings-deny-unknown-fields = ["foundations-macros?/settings-deny-unknown-fields"]
```
inikulin, line 59: Let's make it a macro parameter instead, e.g. `#[settings(deny_unknown_fields = false)]`. With `true` being default in 4.0.0

inikulin (changes requested):

foundations/Cargo.toml
OmegaJak: Done. My initial opposition to this approach was thinking that users of intermediate dependencies like Oxy wouldn't have a way to opt out for shared settings structs. But, in that case, that intermediate dependency can add its own feature flag and use it for the shared settings structs like so:
```
#[cfg_attr(
    feature = "settings-deny-unknown-fields",
    settings(deny_unknown_fields = false)
)]
#[cfg_attr(not(feature = "settings-deny-unknown-fields"), settings)]
```

foundations-macros/src/settings.rs
OmegaJak: Not necessary now that there's no feature flag being used :+1:

foundations/src/settings/mod.rs
```diff
@@ -310,6 +310,23 @@ use std::path::Path;
 /// }
 /// ```
 ///
+/// # Deny unknown fields
+///
+/// By default, this macro will automatically annotate types with `#[serde(deny_unknown_fields)]`.
+/// This means any keys in the config file that don't map directly to fields in the type will
+/// lead to a panic on deserialization.
```
inikulin, line 317: Not panic, but error. We should probably also explain motivation here: that it's usually desirable for services to use up to date configuration scheme and it's better for services to fail early on a bootstrap stage rather than run misconfigured.

inikulin (changes requested):

foundations/src/settings/mod.rs
OmegaJak: Good points, thank you.

inikulin (approved):

PR #52, opened 2024-07-09
ZTC-1478: rework PR #47 in light of the fact that the next release we do is going to have to be a major version bump anyway, no matter what

bcottrell8738:
I did PR #47 in the way that I did it, because I was trying to keep the externally visible API the same. There was a public function, slog_logger(), which promised to return an `Arc<RwLock<Logger>>`. So that meant we had to *have* an  `Arc<RwLock<Logger>>` lying around. This constrained the design and made things grosser than they might otherwise have been.

It has been pointed out that we're already in the situation where the next release we do is going to need to be a major version bump, *anyway*. So we might as well revisit this, and do it the less-gross way that (slightly) breaks API compatibility.

We can minimize the extent to which we break API compatibility by having slog_logger() return a `Arc<RwLock<impl Deref<Target = Logger>>>`. This is nearly the same thing as it used to be.

This PR consists of two commits; one that reverts PR #47, and one that puts the functionality back in the new way. If you are reviewing it, you can either review the PR as a whole (which will show you the diff vs current main branch) or you can review only the second of the two commits (which will show you the diff vs how things looked prior to PR #47). I have tried to keep both sets of aesthetic considerations in mind while preparing this PR.

### DIFF
--- a/foundations/src/telemetry/log/internal.rs
+++ b/foundations/src/telemetry/log/internal.rs
@@ -2,90 +2,57 @@ use super::init::LogHarness;
 use crate::telemetry::scope::Scope;
 use slog::{Logger, OwnedKV, SendSyncRefUnwindSafeKV};
 use std::ops::Deref;
-use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::Arc;
 
-pub const MAX_LOG_GENERATION: u32 = 1000;
-pub const EXCEEDED_MAX_LOG_GENERATION_ERROR: &str = "foundations::telemetry::log: maximum logger generation exceeded (are add_fields! or set_verbosity being called in a loop?)";
-
-// The slog_logger() function exposes Arc<RwLock<Logger>> as a public
-// interface, so we can't store the generation number where it would make the
-// most sense to store it, which would be inside the RwLock, right next to
-// the Logger.
-//
-// Doing it this way means we lose the enforced atomicity between updating the
-// Logger and updating the generation number. In general, we try to only read
-// or write the generation number while we hold the appropriate lock on the
-// Logger. It is not mission-critical that we do this. The generation number
-// is just a guard rail to prevent us from going wild. Even if there is some
-// race condition and we're off by one or two, it won't affect its ability to
-// do its job.
-//
 // NOTE: we intentionally use a lock without poisoning here to not
 // panic the threads if they just share telemetry with failed thread.
-pub(crate) type SharedLog = Arc<LogContext>;
+pub(crate) type SharedLog = Arc<parking_lot::RwLock<LoggerWithKvNestingTracking>>;
 
-#[derive(Debug)]
-pub struct LogContext {
-    // The logger itself. This is the most important part of this struct.
-    pub logger: Arc<parking_lot::RwLock<Logger>>,
+#[derive(Debug, Clone)]
+pub struct LoggerWithKvNestingTracking {
+    // The logger itself. This is the most important part of this struct. (We implement Deref to
+    // let you go straight to this field, in contexts where you need a &Logger)
+    pub(crate) inner: Logger,
 
-    // Generation number, incremented every time the logger is replaced with
-    // a child of itself. Accuracy is not critical as this is only used as a
-    // safety check. Nevertheless, when possible, please be holding the
-    // logger's lock when you read or write the generation number.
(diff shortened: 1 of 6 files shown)


foundations/src/telemetry/log/internal.rs
```diff
@@ -2,93 +2,59 @@ use super::init::LogHarness;
...
-        let mut logger_lock = self.logger.write();
-        let generation = self.generation.fetch_add(1, Ordering::SeqCst) + 1;
-        let logger = f(logger_lock.deref());
-        *logger_lock = logger;
-
-        if generation >= MAX_LOG_GENERATION {
-            panic!("{}", EXCEEDED_MAX_LOG_GENERATION_ERROR);
+        let logger = f(&self.inner);
+        self.inner = logger;
+        self.nesting_level += 1;
```
inikulin, line 38: Style nit: separate variable declarations and `if` statement from the rest of the code with blank lines

foundations/src/telemetry/log/internal.rs
```diff
@@ -2,93 +2,59 @@ use super::init::LogHarness;
...
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        &self.inner
     }
 }
 
+// NOTE: we intentionally use a lock without poisoning here to not
+// panic the threads if they just share telemetry with failed thread.
+pub(crate) type SharedLog = Arc<parking_lot::RwLock<LoggerWithNestingTracking>>;
```
inikulin, line 56: Style nit: move type declarations above all the structure declarations

foundations/src/telemetry/log/internal.rs
```diff
@@ -2,93 +2,59 @@ use super::init::LogHarness;
...
         F: FnOnce(&Logger) -> Logger,
     {
-        let mut logger_lock = self.logger.write();
-        let generation = self.generation.fetch_add(1, Ordering::SeqCst) + 1;
-        let logger = f(logger_lock.deref());
-        *logger_lock = logger;
-
-        if generation >= MAX_LOG_GENERATION {
-            panic!("{}", EXCEEDED_MAX_LOG_GENERATION_ERROR);
+        let logger = f(&self.inner);
```
inikulin, line 36: Having  a function named update is a bit vague and susceptible to misuse - updating inner log not necessary means that the nesting level will increase. Also, it has a bit unconventional argument for updating that would usually imply a mut reference to a property that supposed to be updated. I can't really come up with a good name for this function, it's really something like `replace_inner_with_new_log_with_nested_kv` and it's too verbose. It's a good sign that we probably should have the function at all - let's just update the `inner` field on a callsite and then introduce `inc_kv_nesting_level` fn that will increment nesting and do the assertion.

foundations/src/telemetry/log/internal.rs
```diff
@@ -2,93 +2,59 @@ use super::init::LogHarness;
...
+        self.inner = logger;
+        self.nesting_level += 1;
+        if self.nesting_level >= Self::MAX_LOG_NESTING {
```
inikulin, line 39: Can be `assert!(self.nesting_level < Self::MAX_LOG_NESTING, Self::EXCEEDED_MAX_LOG_NESTING_ERROR)`.

inikulin (changes requested):

foundations/src/telemetry/log/internal.rs
bcottrell8738: Having all "replace logger with a child of itself" updates go through a single code path makes it harder to "miss one". So I felt comfortable with the unit tests only testing add_fields!, because I trusted that if add_fields! enforced the limit, so would set_verbosity().

Now that we don't have a single update() method anymore, I can't implicitly trust that anymore. But I don't want to bloat two unit tests into four, either.

So what I'm doing is I'm having each of the unit tests do approximately MAX/2 add_fields!, and approximately MAX/2 set_verbosity(). (One unit test does slightly under, the other unit test does slightly over)

I think this is sufficient, because if we missed the check in one of the two places, then the unit test that does 501 add_fields! and 501 set_verbosity() would only increment the counter 501 times, not 1002 times. So it wouldn't panic, and the test case would fail.

So I think this is good.

nmldiegues (approved):

inikulin (approved):

PR #54, opened 2024-07-17
ZTC-1648: Avoid heap profiling crash by eagerly starting long-lived profiling thread

### LINKED ISSUE #46: Consider removing seccomp setup in memory_profiler
It's unusual to handle the library responsiblity to setup seccomp by itself.  The seccomp setup will require some additional syscalls, such as `prctl`, and if the process has setup seccomp already and do not allow those syscalls anymore, then it will fail.
And if the main process didn't setup seccomp, then it does not make sense for memory_profile to setup seccomp itself, because obviously we didn't care.

I would propose to remove the seccomp setup in memory_profiler.

OmegaJak:
Closes #46

Rather than fully removing seccomp setup in the memory profiler, which would require adding new syscalls to the whole application's configuration, the memory profiler now spawns a long-lived thread during startup, before global seccomp is initialized. This thread no longer sets up seccomp for itself.


foundations/src/telemetry/memory_profiler.rs
```diff
@@ -56,12 +59,12 @@ struct Seal;
...
 /// [jemalloc]: https://github.com/jemalloc/jemalloc
-#[derive(Copy, Clone)]
+#[derive(Clone)]
 pub struct MemoryProfiler {
     _seal: Seal,
 
-    #[cfg(feature = "security")]
-    sandbox_profiling_syscalls: bool,
+    request_heap_profile: mpsc::Sender<oneshot::Sender<anyhow::Result<String>>>,
+    _heap_profiling_thread_handle: Arc<Mutex<JoinHandle<()>>>,
```
OmegaJak, line 67: I wasn't sure what to do with this JoinHandle, since we anticipate the thread living for the whole life of the application.

Should we just drop the handle and detach the thread? The thread will still die once all mpsc::Senders are dropped.

foundations/src/telemetry/memory_profiler.rs
```diff
@@ -182,26 +189,67 @@ fn init_profiler(settings: &MemoryProfilerSettings) -> BootstrapResult<Option<Me
...
+
+fn heap_profile_thread(
+    receive_request: mpsc::Receiver<oneshot::Sender<anyhow::Result<String>>>,
+    #[cfg(feature = "security")] setup_complete: mpsc::Sender<anyhow::Result<()>>,
+    sandbox_profiling_syscalls: bool,
+) {
+    #[cfg(feature = "security")]
+    if sandbox_profiling_syscalls {
+        if let Err(_) = setup_complete.send(sandbox_jemalloc_syscalls()) {
+            return;
```
OmegaJak, line 226: Should we log something here? I didn't see much of a point as this can only happen when the main/parent thread drops the `Receiver` corresponding to this `Sender`, which should only happen in error scenarios, where something else is already going wrong.

Same question for the loop below

foundations/src/telemetry/memory_profiler.rs
```diff
@@ -256,6 +306,30 @@ mod tests {
         .is_err());
     }
 
+    #[tokio::test]
+    async fn profile_heap_with_profiling_sandboxed_after_previous_seccomp_init() {
```
OmegaJak, line 310: This test failed before this PR's changes

foundations/src/telemetry/memory_profiler.rs
```diff
@@ -256,6 +302,30 @@ mod tests {
...
+        let profiler = MemoryProfiler::get_or_init_with(&MemoryProfilerSettings {
```
inikulin, line 307: As far as I can tell this will still fail as `MemoryProfiler` will usually be initialised in production on a first request to telemetry server for the a heap profile and that happens when the whole app is spin up and seccomp is initialised on the main thread.

It seems the way to solve that is to start the profiling thread in `telemetry::init` that is recommended to be called before seccomp init on the main thread (though, it's not reflected in docs, but shown in the example - probably we should update the docs here). 

Task sender for the thread will be stored in a global var (like the profiler itself), so once we have profiler initialized, it can tell the thread to collect a profile via a global handle. Additionally, we can put the sender under a mutex, this way we can remove `PROFILING_IN_PROGRESS_LOCK`.

inikulin (changes requested):

foundations/src/telemetry/memory_profiler.rs
OmegaJak: With my changes, the `MemoryProfiler` is initialized when `telemetry::server::init` is [called](https://github.com/cloudflare/foundations/pull/54/files#diff-905ca2b1e6764e2f0cc3b222abef579e7eb319275a9e6a1161a5fd749092cdefR54), which is called by `telemetry::init`. I'll make sure the recommendation to call `telemetry::init` before seccomp setup is in the docs.

Regarding the locking -- if the profiler itself is stored in a global variable, and the only way to get the profiler is through the global variable, I'm not sure I see the point of pulling the sender into its own global variable rather than letting it continue living inside the profiler.

As far as the `PROFILING_IN_PROGRESS_LOCK` goes, looking at it now, I actually don't think we need that any more. If multiple requests come in at the same time, they'll simply queue up and each be processed in turn by the profiling thread's loop. I'll look into removing that altogether.

OmegaJak: I've added docs and removed the `PROFILING_IN_PROGRESS_LOCK`. As a sanity check, I used apache bench to hit a local version of a service using this branch, making 10k heap profile requests with a concurrency level of 100. High concurrency meant high response times of course, but there were no failures.

<details>

<summary>Apache Bench Output</summary>

```
$ ab -n 10000 -c 100 -l localhost:7800/pprof/heap
This is ApacheBench, Version 2.3 <$Revision: 1879490 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:        
Server Hostname:        localhost
Server Port:            7800

Document Path:          /pprof/heap
Document Length:        Variable

Concurrency Level:      100
Time taken for tests:   1.236 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      88918063 bytes
HTML transferred:       87658063 bytes
Requests per second:    8091.14 [#/sec] (mean)
Time per request:       12.359 [ms] (mean)
Time per request:       0.124 [ms] (mean, across all concurrent requests)
Transfer rate:          70258.63 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       1
Processing:     1   12   3.1     11      20
Waiting:        1   12   3.1     11      20
Total:          2   12   3.1     11      20

Percentage of the requests served within a certain time (ms)
  50%     11
  66%     13
  75%     14
  80%     15
  90%     17
  95%     19
  98%     19
  99%     19
 100%     20 (longest request)

```

</details>

foundations/src/telemetry/memory_profiler.rs
```diff
@@ -1,24 +1,25 @@
...
 use std::io::Read;
 use std::os::raw::c_char;
-use std::thread;
+use std::sync::mpsc::{self};
+use std::sync::{Arc, Mutex};
+use std::thread::{spawn, JoinHandle};
 use tempfile::NamedTempFile;
-use tokio::sync::Mutex as AsyncMutex;
-use tokio::task::spawn_blocking;
+use tokio::sync::oneshot;
```
nmldiegues, line 15: Do we really require tokio's `oneshot` or could we use `std` one?

foundations/src/telemetry/memory_profiler.rs
nmldiegues: I can only think of the use case of possibly observing a catastrophic failure of that thread (such as a panic) that causes it to end?

could it be useful to pass the handle to a spawned tokio task that joins the handle to observe the result of the thread?  that way if it fails unexpectedly we can log an error and it becomes observable, whereas in the thread itself we may not be able to handle unexpected panics?

nmldiegues: Side-note:  why do you have a Mutex around the handle?

foundations/src/telemetry/memory_profiler.rs
```diff
@@ -182,26 +182,63 @@ fn init_profiler(settings: &MemoryProfilerSettings) -> BootstrapResult<Option<Me
...
+fn receive_profiling_thread_setup_msg(
+    setup_complete_receiver: mpsc::Receiver<anyhow::Result<()>>,
+) -> anyhow::Result<()> {
+    match setup_complete_receiver.recv() {
+        Ok(Ok(())) => {}
+        Ok(Err(setup_err)) => {
+            return Err(setup_err);
+        }
+        Err(std::sync::mpsc::RecvError) => {
+            bail!("Profiling thread disconnected before finishing setup")
```
nmldiegues, line 200: will this show the actual error?

foundations/src/telemetry/memory_profiler.rs
nmldiegues: Yeah if the main thread is not there to receive from this channel, then I agree something else must have already happened that would be observed.

nmldiegues: Worth dropping a comment in this function to explain that though

foundations/src/telemetry/memory_profiler.rs
OmegaJak: Having something else join the thread feels rather high-cost for a (hopefully) niche scenario, but it's probably better to have it in case the profiling thread does happen to panic. Unfortunately, it's a std `JoinHandle` which means joining it is a blocking operation. So this other watchdog task I think would actually have to be another std thread, not a tokio task. That, or we do some kind of polling on whether the heap profiling thread is finished. That doesn't seem great either. I'll see how it looks to have another watchdog thread.

The Mutex is superfluous in the current impl, and will be with the watchdog as well, I'll remove it

foundations/src/telemetry/memory_profiler.rs
OmegaJak: Not sure exactly what you're referring to - there isn't a `oneshot` in `std`. There's the [oneshot crate](https://crates.io/crates/oneshot), otherwise, as far as I know, `std` just has its general `mpsc`.

Regardless, I think tokio's `oneshot` is the perfect construct for what we want. Its semantics are preferable over an `mpsc` since we always expect one request to produce exactly one response, which its interface captures perfectly. And I think waiting on an async `Receiver` is preferable over a blocking/synchronous one since the actual work the heap profiler is doing is truly on a separate `std::thread`, so the thread asynchronously handling the HTTP request can theoretically do other things while waiting.

foundations/src/telemetry/memory_profiler.rs
OmegaJak: As far as I can tell, the Error returned by this method will bubble all the way up the call stack to where the telemetry server is initialized

foundations/src/telemetry/memory_profiler.rs
OmegaJak: I pursued a few avenues to join the profiling thread (using another thread, doing it when we notice sending to it fails), but they were all rather hairy. And, I believe, unnecessary. When a detached thread panics it still invokes the global panic hook, meaning it will write to stderr. In Oxy, we also extend the default panic hook by writing it to our logs with context, so a panic in the profiling thread will be logged. This seems sufficient to me.

foundations/src/telemetry/memory_profiler.rs
OmegaJak: Added comment

foundations/src/telemetry/memory_profiler.rs
nmldiegues: Good point, we just got to detach it.

foundations/src/telemetry/memory_profiler.rs
nmldiegues: > there isn't a oneshot in std.

Oh, I thought there would be, I honestly didn't bother checking. In that case, nevermind.

nmldiegues (approved):

foundations/src/telemetry/memory_profiler.rs
inikulin: Even though having multiple profiles can be fine in artificial benchmarks, we don't want to run multiple profiles at a time on a heavily loaded server as it: 1) can have effect on the performance of the server; 2) due to 1 can produce skewed results for other profiles, e.g. timing discrepancy can cause pacing issues and introduce additional allocations; 3) due to the way how our profiling pipeline is built (i.e. profiles are available globally once they collected) we want to avoid doing the same job multiple times and just allow others grab a profile that was already collected by someone else at the time period when they requested theirs.

inikulin: > With my changes, the MemoryProfiler is initialized when telemetry::server::init is [called](https://github.com/cloudflare/foundations/pull/54/files#diff-905ca2b1e6764e2f0cc3b222abef579e7eb319275a9e6a1161a5fd749092cdefR54),

I see. Let's also add docs for those who might use the profiler programmatically outside of telemetry server that if they have seccomp enabled it's recommended to init the profiler before that.

inikulin (changes requested):

foundations/src/telemetry/memory_profiler.rs
OmegaJak: > we don't want to run multiple profiles at a time on a heavily loaded server

We don't run multiple profiles at a time. Requests enter into a queue which the profiling thread processes one at a time.

> due to the way how our profiling pipeline is built (i.e. profiles are available globally once they collected) we want to avoid doing the same job multiple times and just allow others grab a profile that was already collected by someone else at the time period when they requested theirs

Sure, we could do something like "if we receive profiling request A, and then request B comes in while we're still gathering A's, then return to B what we returned to A". But, that just seems like a premature optimization with unneeded complexity to me.

If you're opposed to the queuing functionality and want to match how it worked before my changes, we could avoid it by reintroducing a lock/mutex around the sender and returning an error if we can't immediately acquire the lock. While that does avoid possible server overload due to profiling requests with a sort of rate limiting of "only one at a time", it feels like an unnecessary limiting of the server's functionality. If we do that, and user B requests a profile while user A's is still running, user B will just get an error and have to try again, instead of (as it currently is) getting their profile result a few milliseconds later when A's is done.

foundations/src/telemetry/memory_profiler.rs
```diff
@@ -56,12 +55,11 @@ struct Seal;
...
 ///
 /// [jemalloc]: https://github.com/jemalloc/jemalloc
-#[derive(Copy, Clone)]
+#[derive(Clone)]
 pub struct MemoryProfiler {
     _seal: Seal,
 
-    #[cfg(feature = "security")]
-    sandbox_profiling_syscalls: bool,
+    request_heap_profile: mpsc::Sender<oneshot::Sender<anyhow::Result<String>>>,
```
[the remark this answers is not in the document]
OmegaJak: Good point, thank you! Removed.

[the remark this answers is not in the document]
OmegaJak: Done

foundations/src/telemetry/memory_profiler.rs
```diff
@@ -169,6 +148,26 @@ fn init_profiler(settings: &MemoryProfilerSettings) -> BootstrapResult<Option<Me
...
     }
 
+    let (request_sender, request_receiver) = mpsc::channel();
+
+    #[cfg(feature = "security")]
+    let (setup_complete_sender, setup_complete_receiver) = mpsc::channel();
+
+    #[cfg(feature = "security")]
+    let sandbox_profiling_syscalls = settings.sandbox_profiling_syscalls;
+    std::thread::spawn(move || {
```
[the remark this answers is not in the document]
OmegaJak: Done

foundations/src/telemetry/memory_profiler.rs
inikulin: > We don't run multiple profiles at a time. Requests enter into a queue which the profiling thread processes one at a time.

Don't really want to allow queuing profiles: we either need to have a sophisticated way to control and monitor the queue (e.g. metrics for queue, duration, a way to cancel the queue, etc.). Some of the applications are extremely loaded and hypersensitive to performance changes and at the same time anyone really can request a profile. And I imagine people erroneously queuing profiles multiple times. And the only way we can abort such a queue is only via  a service restart. Then, again, if people request profiles, they probably run a certain experiment: making requests/connections, etc. So, the timing is important, running profile without a timing guarantee of when it will be run is not very useful. To make the matter worse, we don't even give any feedback signal to the requester at the moment whether profile will be run immediately or queued. 

So, tl;dr, yes, let's re-introduce the lock.

foundations/src/telemetry/memory_profiler.rs
```diff
@@ -212,50 +192,50 @@ fn collect_heap_profile() -> Result<String> {
...
 mod tests {
     use super::*;
-    use crate::telemetry::settings::MemoryProfilerSettings;
+    use crate::{
+        security::{
+            allow_list,
+            common_syscall_allow_lists::{ASYNC, SERVICE_BASICS},
+            enable_syscall_sandboxing, ViolationAction,
+        },
+        telemetry::settings::MemoryProfilerSettings,
```
inikulin, line 204: Nit: don't use nested imports. Instead, introduce a separate `use` for each non-leaf path, e.g.:
```rust
use crate::security::{allow_list, enable_syscall_sandboxing, ViolationAction};
use crate::security::common_syscall_allow_lists::{ASYNC, SERVICE_BASICS};
...
```

inikulin (changes requested):

foundations/src/telemetry/memory_profiler.rs
OmegaJak: Lock re-introduced.

foundations/src/telemetry/memory_profiler.rs
OmegaJak: Ah the joys of auto importing. Too bad rustfmt's [imports_granularity = "Module"](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=Module) config isn't stabilized yet

OmegaJak:
CI failure should be fixed with #57

inikulin (approved):

Issue #39, opened 2024-03-21
Change LogScope to use an _inner field instead of allow(dead_code)


nox:
https://github.com/cloudflare/foundations/pull/38

PR #55, opened 2024-07-18
Telemetry API improvements

inikulin:
- Provide API to transform `SpanScope` into `TelemetryContext`. That allows ergonomically apply span scopes returned by `start_trace` to futures. In addition, it allows us to get rid of `TelemetryContext::apply_with_tracing_span` and alike
- Add telemetry instrumentation support for `!Send` futures
- Remove usages of `#[allow(dead_code)]` (closes https://github.com/cloudflare/foundations/issues/39)

### DIFF
--- a/foundations-macros/src/span_fn.rs
+++ b/foundations-macros/src/span_fn.rs
@@ -41,11 +41,18 @@ impl Parse for SpanName {
 struct Options {
     #[darling(default = "Options::default_crate_path")]
     crate_path: Path,
+
+    #[darling(default = "Options::default_async_local")]
+    async_local: bool,
 }
 
 impl Options {
     fn default_crate_path() -> Path {
         parse_quote!(::foundations)
+    }
+
+    fn default_async_local() -> bool {
+        false
     }
 }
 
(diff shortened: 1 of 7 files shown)


foundations/src/telemetry/telemetry_context.rs
nmldiegues, line 1: /rant

How hard would be for Github to properly diff this against the code where this was moved from. Oh man...

foundations-macros/src/span_fn.rs
```diff
@@ -41,12 +41,19 @@ impl Parse for SpanName {
 struct Options {
     #[darling(default = "Options::default_crate_path")]
     crate_path: Path,
+
+    #[darling(default = "Options::default_async_local")]
+    async_local: bool,
```
nmldiegues, line 46: is this commonly known to carry the meaning/semantics you want to?

or would it be better to call it `not_send`?

nmldiegues (approved):

foundations-macros/src/span_fn.rs
inikulin: As far as I can tell, yes:
- https://docs.rs/tokio/latest/tokio/task/struct.LocalSet.html (search for "local futures")
- https://docs.rs/futures/latest/futures/future/type.LocalBoxFuture.html

cbranch (approved):

PR #59, opened 2024-07-30
Add ability for logging to record tracing span log entries

### LINKED ISSUE #58: Add option to collect log lines emitted in traced contexts in span logs
Example use case:

An HTTP service, for which some requests are sampled through distributed tracing, whilst its syslogs are sampled/rate limited in collection - so for a traced request with a failure, it's possible the log information from that error has been dropped, or if it is, exists in some separate system.

Instead, as foundations already provides both a `logging` and `tracing` module, the logging module could be configurable to additionally send log entries over to the tracing module. Unlike other drains and pipelines, this would not be heavily filtered and sampled - so the logs that are most useful for the requests we have traced information for are the ones we know we will have retained, right where we need to see them.

LloydW93:
In large environments with separate logging and tracing pipelines, identifying pertinent log information for a trace can involve searching in an entirely different store, which may be subject to its own sampling and rate limiting. For these situations, the log lines generate specifically whilst the traced action is executed are particularly high value in providing additional context.

This change adds trace emission options for the logging module, which allow enabling emitting logs from traces at a separately specified verbosity level, which can be changed the same was as we can with set_verbosity.

To fully specify log fields, add_span_log_fn is added as a macro to the tracing module, at a cost of losing the full abstraction over rustracing.

Fixes #58

### DIFF
--- a/foundations/src/telemetry/log/init.rs
+++ b/foundations/src/telemetry/log/init.rs
@@ -2,6 +2,10 @@ use super::field_dedup::FieldDedupFilterFactory;
 use super::field_filtering::FieldFilteringDrain;
 use super::field_redact::FieldRedactFilterFactory;
 use super::internal::{LoggerWithKvNestingTracking, SharedLog};
+#[cfg(feature = "tracing")]
+use super::tracing_drain::{TracingDrain, TracingFilteringDrain};
+#[cfg(feature = "tracing")]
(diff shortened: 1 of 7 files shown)


inikulin:
@LloydW93 I'm not sure I understand the motivation behind #58 (see comment there) and what this PR tries to do

PR #60, opened 2024-08-01
OXY-1404: Avoid crashes resulting from double seccomp initialization

OmegaJak:
If a service using foundations accidentally initializes seccomp in a thread that already seccomp initialized, the seccomp violation (depending on configuration) may be violated and crash the process.

An easy way to enter into this scenario is by having a hook in [tokio::runtime::Builder::on_thread_start](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.on_thread_start) that initializes seccomp for threads that are initialized before the main thread. In this scenario, any threads created from tokio threads (say, inside a tokio task) may lead to a seccomp violation.

This PR uses the syscall `prctl(PR_GET_SECCOMP)` ([PR_GET_SECCOMP](https://linuxman7.org/linux/man-pages/man2/PR_GET_SECCOMP.2const.html)) to find out definitively whether seccomp is already enabled for the current thread. If it is, then the default behavior is to not initialize seccomp again, thus avoiding what may have been a crash otherwise.

### DIFF
--- a/foundations/src/security/mod.rs
+++ b/foundations/src/security/mod.rs
@@ -48,8 +48,10 @@ mod sys {
 }
 
 use self::internal::RawRule;
-use crate::BootstrapResult;
-use anyhow::bail;
+use crate::{BootstrapError, BootstrapResult};
+use anyhow::{anyhow, bail};
+use std::fmt::Display;
+use sys::PR_GET_SECCOMP;
 
 pub use self::syscalls::Syscall;
 
@@ -97,6 +99,46 @@ impl ArgCmpValue {
 impl<T: Into<u64>> From<T> for ArgCmpValue {
     fn from(val: T) -> Self {
         Self(val.into())
+    }
+}
+
+/// Errors that can be produced when initializing sandboxing using
+/// [`enable_syscall_sandboxing`]
+#[derive(Debug)]
+pub enum SandboxingInitializationError {
+    /// Sandboxing has already been initialized on the current thread
+    /// with the given [`SeccompMode`]
+    AlreadyInitialized(SeccompMode),
+
+    /// Some other error occurred during initialization
+    Other(BootstrapError),
+}
+
+impl Display for SandboxingInitializationError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            SandboxingInitializationError::AlreadyInitialized(mode) => write!(
+                f,
+                "seccomp has already been initialized and is in mode {:?}",
+                mode
+            ),
+            SandboxingInitializationError::Other(e) => e.fmt(f),
+        }
+    }
+}
+
+impl From<BootstrapError> for SandboxingInitializationError {
+    fn from(err: BootstrapError) -> Self {
+        SandboxingInitializationError::Other(err)
+    }
+}
+
+impl From<SandboxingInitializationError> for BootstrapError {
+    fn from(value: SandboxingInitializationError) -> Self {
+        match value {
+            SandboxingInitializationError::Other(e) => e,
+            e @ SandboxingInitializationError::AlreadyInitialized(_) => anyhow!("{}", e),
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


OmegaJak:
There are lots of ways to implement this from an API perspective, I'm totally open to alternatives. We could leave `enable_syscall_sandboxing` untouched and just provide the new `get_current_thread_seccomp_mode()` method, expecting callers to check this method first if they care. But that felt to me like leaving a footgun in place for the standard path of using `enable_syscall_sandboxing`.

We could also not provide the `enable_syscall_sandboxing_ignore_existing` method, but it's technically possible that there are currently applications that initialize seccomp twice, which that would break.

I'm happy to implement any of those alternative approaches.

foundations/src/security/mod.rs
```diff
@@ -348,6 +370,27 @@ pub enum Rule {
 pub fn enable_syscall_sandboxing(
     violation_action: ViolationAction,
     exception_rules: &Vec<Rule>,
+) -> BootstrapResult<()> {
+    let current_mode = get_current_thread_seccomp_mode()?;
+    if current_mode != SeccompMode::None {
+        return Ok(());
```
cbranch, line 376: Silently succeeding seems like the wrong method here, if the exception list is different to what's actually applied you could cause unexpected failures later on. I would feel happier with an error that we can detect and report in a more sensible manner (and/or the app can detect this is needed)

OmegaJak: Yeah, I debated that myself. It now returns an Error. Since we need the caller to be able to distinguish this from other things going wrong during initialization, I introduced an error type for this. It should make for nice discoverability when using this API for the first time, and still gives callers the choice of what to do (whereas previously seccomp most likely would have just terminated the process).

OmegaJak:
This now has a dependency on #54 - it removed a bit of code that needs error type conversion now. So the compilation will fail until that PR is merged.

edoerpani (approved):

foundations/src/security/mod.rs
```diff
@@ -47,9 +47,12 @@ mod sys {
     include!(concat!(env!("OUT_DIR"), "/security_sys.rs"));
 }
 
+use std::fmt::Display;
+
```
nmldiegues, line 51: remove blank space and reorder

foundations/src/security/mod.rs
```diff
@@ -388,6 +476,31 @@ pub fn enable_syscall_sandboxing(
...
+///
+/// allow_list! {
+///    pub static MY_ALLOW_LIST = [
+///        ..RUST_BASICS
+///    ]
+/// }
+/// ```
+///
+/// [prctl(PR_GET_SECCOMP)]: https://linuxman7.org/linux/man-pages/man2/PR_GET_SECCOMP.2const.html
+pub fn get_current_thread_seccomp_mode() -> BootstrapResult<SeccompMode> {
```
nmldiegues, line 495: Does this need to be `pub`?

nmldiegues (approved):

foundations/src/security/mod.rs
OmegaJak: Done

foundations/src/security/mod.rs
OmegaJak: Doesn't need to be, no. But I figured since we had it, that it could be useful to users of foundations - though I'm not really sure how.

I guess this comes down to "expose it because we have it" vs "keep it private to minimize public surface area unless definitely needed". I lean towards the former in this case, but don't feel strongly either way, so if you'd prefer it be made private I'd be happy to do so.

nmldiegues: I usually only expose things when we already have use cases in mind for it

OmegaJak: Yeah that's fair. I've removed `pub` from its definition and moved it down. I left the docstring since it was already written even though it's not "required" any more

cbranch (approved):

PR #63, opened 2024-08-20
chore: add passive sampling option to tracing

lbarthon:
When we have many services in the same flow, we want to defer the sampling logic to the parent service (which will defer to the root service in the end).

The PassiveSampler is really good for that, as it will only sample traces that have one external reference. This assumes that the parent service will send the trace info to stitch, if we should keep it, and won't, if we should discard it. This is what some of our services do already today, and it would really help us onboard to foundations!

This needs https://github.com/cloudflare/rustracing/pull/2 merged & `cf-rustracing` bumped in order to compile and work fine.

### DIFF
--- a/foundations/src/telemetry/settings/tracing.rs
+++ b/foundations/src/telemetry/settings/tracing.rs
@@ -24,15 +24,8 @@ pub struct TracingSettings {
     /// The output for the collected traces.
     pub output: TracesOutput,
 
-    /// Sampling ratio.
-    ///
-    /// This can be any fractional value between `0.0` and `1.0`.
-    /// Where `1.0` means "sample everything", and `0.0` means "don't sample anything".
-    #[serde(default = "TracingSettings::default_sampling_ratio")]
-    pub sampling_ratio: f64,
-
-    /// Settings for rate limiting emission of traces
-    pub rate_limit: RateLimitingSettings,
+    /// The strategy used to sample traces.
+    pub sampling_strategy: SamplingStrategy,
 }
 
 #[cfg(not(feature = "settings"))]
@@ -41,8 +34,7 @@ impl Default for TracingSettings {
         Self {
             enabled: TracingSettings::default_enabled(),
             output: Default::default(),
-            sampling_ratio: TracingSettings::default_sampling_ratio(),
-            rate_limit: Default::default(),
+            sampling_strategy: Default::default(),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


inikulin (changes requested):

foundations/src/telemetry/tracing/init.rs
```diff
@@ -60,9 +61,13 @@ impl TracingHarness {
...
-    Ok(Tracer::new(RateLimitingProbabilisticSampler::new(
-        settings,
-    )?))
+    let sampler = match &settings.sampling_strategy {
+        SamplingStrategy::Passive => PassiveSampler.boxed(),
+        SamplingStrategy::Active(settings) => {
+            RateLimitingProbabilisticSampler::new(settings)?.boxed()
+        }
+    };
+    Ok(Tracer::new(sampler))
```
inikulin, line 70: Nit: blank line before this expression

foundations/src/telemetry/settings/tracing.rs
```diff
@@ -125,3 +113,56 @@ impl JaegerThriftUdpOutputSettings {
...
+#[cfg_attr(
+    feature = "settings",
+    settings(crate_path = "crate", impl_default = false)
+)]
+#[cfg_attr(not(feature = "settings"), derive(Clone, Debug, serde::Deserialize))]
+pub enum SamplingStrategy {
+    /// This only samples traces which have one or more references.
+    ///
+    /// Backed by [cf_rustracing::sampler::PassiveSampler]
+    Passive,
```
inikulin, line 158: Could we elaborate here a bit more about active vs passive sampling in doc comments for these options?

foundations/tests/tracing.rs
```diff
@@ -18,10 +20,13 @@ fn test_rate_limiter(mut ctx: TestTelemetryContext) {
     assert_eq!(ctx.traces(Default::default()).len(), 10);
 
     ctx.set_tracing_settings(TracingSettings {
-        rate_limit: RateLimitingSettings {
-            enabled: true,
-            max_events_per_second: 5,
-        },
+        sampling_strategy: SamplingStrategy::Active(ActiveSamplingSettings {
```
inikulin, line 23: It would be nice to have some basic test for passive sampling

lbarthon: Fixed the current test, as it wasn't behaving as expected, and added a simple one for passive sampling

PR #65, opened 2024-09-20
Provide foundations::telemetry::metrics::with_registry()

bobrik:
Currently there is no way to enable third party libraries to register metrics, which is not great. With this new method those libraries can get a `Registry` to register their metrics when they initialize, while still giving the end user control over prefixes and external labels.

### DIFF
--- a/foundations/src/telemetry/metrics/mod.rs
+++ b/foundations/src/telemetry/metrics/mod.rs
@@ -12,6 +12,7 @@
 use super::settings::MetricsSettings;
 use crate::Result;
 use prometheus::{Encoder, TextEncoder};
+use prometheus_client::registry::Registry;
 use serde::Serialize;
 use std::any::TypeId;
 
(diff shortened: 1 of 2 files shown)


foundations/src/telemetry/metrics/mod.rs
```diff
@@ -369,3 +370,24 @@ impl MetricConstructor<TimeHistogram> for HistogramBuilder {
         TimeHistogram::new(self.buckets.iter().cloned())
     }
 }
+
+/// Runs a closure with a prometheus [Registry] to enable registration of external metrics.
+///
```
inikulin, line 375: Could elaborate a bit more in the code comment about possible use cases

inikulin: The example doesn't compile also

bobrik: Updated to address both these things.

bobrik:
#66 is somewhat more awkward, but compatible with any `prometheus_client` crate version. For this PR the dependency version of the third party crate has to match that of `foundations`, which is problematic. In #66 there is no such requirement.

PR #66, opened 2024-09-21
Provide metrics::add_extra_producer() to enable external metrics

bobrik:
This is a more forward and backward compatible solution than https://github.com/cloudflare/foundations/pull/65.


inikulin (approved):

nox (approved):

dqminh:
This still depends on an implicit assumption that foundations and the additional producers produce only in text format which is very stable, so don't think it will change. If either uses different encoding for whatever reasons, then we will produce broken format.

To make this goaway probably still require the same `prometheus_client` as seen with #65

bobrik:
This still needs to be merged. We now have it in v3.4.0, but not in v4.0.0.

PR #70, opened 2024-10-02
Fix libssecomp typo

tobyhede:
`libssecomp` is actually `libseccomp`

It will bug me if I left this out there.

### DIFF
--- a/foundations/build.rs
+++ b/foundations/build.rs
@@ -14,7 +14,7 @@ fn ensure_seccomp_sources_fetched() {
 
     if !src_dir.exists() {
         panic!(
-            "Can't find libssecomp sources. Run `git submodule update --init --recursive`. \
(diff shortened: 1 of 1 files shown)


foundations/build.rs
```diff
@@ -14,7 +14,7 @@ fn ensure_seccomp_sources_fetched() {
 
     if !src_dir.exists() {
         panic!(
-            "Can't find libssecomp sources. Run `git submodule update --init --recursive`. \
+            "Can't find libsseccomp sources. Run `git submodule update --init --recursive`. \
```
nmldiegues, line 17: While you're at it, you may as well fix the other typo? :)

Should be `libseccomp`

tobyhede: 🤦 

The typo fix has a typo. Of course it does.

nmldiegues (approved):

PR #72, opened 2024-10-02
OXY-1404: Revert #60 and instead add ability to pass TSYNC attribute to seccomp

OmegaJak:
While updating Oxy with foundations 4.0, I discovered a flaw with the previous approach: [docker](https://docs.docker.com/engine/security/seccomp/#pass-a-profile-for-a-container), by default, uses seccomp to block 44 system calls. This means that, with the approach used in #60, code calling `enable_syscall_sandboxing` would get an error because seccomp is already active if it happened to be running inside a docker container - even though it could still safely call `enable_syscall_sandboxing`.

As described in #60, the original motivation for that change was because of the use of the [tokio::runtime::Builder::on_thread_start](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.on_thread_start) hook leading to double seccomp initialization. This hook was used to ensure that any threads initialized before seccomp is configured on the main thread also had seccomp active, but it led to a risk of double seccomp initialization leading to a crash.

This PR removes the need for any of that using the seccomp flag `SECCOMP_FILTER_FLAG_TSYNC` ([man seccomp](https://man7.org/linux/man-pages/man2/seccomp.2.html)). This flag synchronizes the seccomp filter state of all threads in the current process. Using this, we can remove the `on_thread_start` hook and call `enable_syscall_sandboxing` exactly once in the whole application, ensuring that:
1. All threads created before `enable_syscall_sandboxing` is called will have seccomp configured thanks to the flag
2. All threads created after `enable_syscall_sandboxing` is called will have seccomp configured as the current thread's seccomp state is inherited by new threads
3. We never violate seccomp by attempting to initialize it twice


OmegaJak:
After external discussion, we realized this would break graceful restarts with [shellflip](https://github.com/cloudflare/shellflip), which is currently relying on the thread used to spawn the child not having seccomp configured.

PR #74, opened 2024-10-03
OXY-1404: Revert most of #60 and expose get_current_thread_seccomp_mode

OmegaJak:
(copied from #72's description) While updating Oxy with foundations 4.0, I discovered a flaw with the previous approach: [docker](https://docs.docker.com/engine/security/seccomp/#pass-a-profile-for-a-container), by default, uses seccomp to block 44 system calls. This means that, with the approach used in https://github.com/cloudflare/foundations/pull/60, code calling enable_syscall_sandboxing would get an error because seccomp is already active if it happened to be running inside a docker container - even though it could still safely call enable_syscall_sandboxing.

As described in https://github.com/cloudflare/foundations/pull/60, the original motivation for that change was because of the use of the [tokio::runtime::Builder::on_thread_start](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.on_thread_start) hook leading to double seccomp initialization. This hook was used to ensure that any threads initialized before seccomp is configured on the main thread also had seccomp active, but it led to a risk of double seccomp initialization leading to a crash.

This PR rolls back most of the changes in #60, instead simply exposing the `get_current_thread_seccomp_mode` method so applications can decide for themselves how to handle these scenarios. In our case, we'll use it to log a warning if we're about to enable seccomp and it's already enabled.

### DIFF
--- a/foundations/src/security/mod.rs
+++ b/foundations/src/security/mod.rs
@@ -48,9 +48,8 @@ mod sys {
 }
 
 use self::internal::RawRule;
-use crate::{BootstrapError, BootstrapResult};
-use anyhow::{anyhow, bail};
-use std::fmt::Display;
+use crate::BootstrapResult;
+use anyhow::bail;
 use sys::PR_GET_SECCOMP;
 
 pub use self::syscalls::Syscall;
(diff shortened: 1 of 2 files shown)


nmldiegues (approved):

foundations/src/security/mod.rs
```diff
@@ -529,7 +460,7 @@ pub fn forbid_x86_64_cpu_cycle_counter() {
 /// ```
 ///
 /// [prctl(PR_GET_SECCOMP)]: https://linuxman7.org/linux/man-pages/man2/PR_GET_SECCOMP.2const.html
-fn get_current_thread_seccomp_mode() -> BootstrapResult<SeccompMode> {
+pub fn get_current_thread_seccomp_mode() -> BootstrapResult<SeccompMode> {
```
inikulin, line 463: The example seems to be incomplete - it doesn't show the usage of the API

inikulin: We also have an unused option in `SeccompMode` which is confusing. Should we instead just have `is_seccomp_enabled_for_thread() -> BootstrapResult<bool>`?

OmegaJak: Good point - I'll fill out the example

Also, yeah, I agree about SeccompMode. I think I did that to try and communicate the quirk that you'll never find out that you're in strict mode, since if you call `prctl` in strict mode, your process will be killed. But I think I can do that well enough in the doc on this method and improve the API as you described, since almost every caller would probably do `matches!(seccomp_mode, SeccompMode::Filter)` or similar to convert its return type to a boolean.

I'll open a new PR addressing those.

OmegaJak: See #75

PR #82, opened 2024-10-22
Use global_queue_depth instead of deprecated injection_queue_depth

bobrik:


bobrik:
Welp, now `minimal-versions` job fails because `global_queue_depth` was only introduced in tokio v1.41.0:

* https://github.com/tokio-rs/tokio/commit/28c9a14a2e4da99842d41240b301022017b2a04a

I'm not sure what the right next step is here.

inikulin:
> Welp, now `minimal-versions` job fails because `global_queue_depth` was only introduced in tokio v1.41.0:
> 
> * [tokio-rs/tokio@28c9a14](https://github.com/tokio-rs/tokio/commit/28c9a14a2e4da99842d41240b301022017b2a04a)
> 
> I'm not sure what the right next step is here.

Need to bump tokio to this version in `Cargo.toml`. @Noah-Kennedy we desperately need docs on min version, could you please contribute it, since you introduced the check?

PR #88, opened 2024-11-08
Properly await for all futures in Future for TelemetryDriver

nox:


foundations/src/telemetry/driver.rs
```diff
@@ -88,25 +87,19 @@ impl Future for TelemetryDriver {
...
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        let mut ready_res = vec![];
-
         #[cfg(feature = "telemetry-server")]
         if let Some(server_fut) = &mut self.server_fut {
-            if let Poll::Ready(res) = Pin::new(server_fut).poll(cx) {
-                ready_res.push(res.map_err(|err| anyhow!(err)));
+            if let Poll::Ready(()) = Pin::new(server_fut).poll(cx)? {
+                self.server_fut = None;
```
inikulin, line 93: doesn't it mean we'll poll the server just once?

nox: No, this checks for `Poll::Ready(())`, so it will be polled until it's ready.

cbranch (approved):

PR #89, opened 2024-11-12
More expressive error for EMSGSIZE from Jaeger reporter

TheJokr:
We observe around 22k `EMSGSIZE` errors per day in production and don't have any information to investigate further. By adding the span name as well as some tag and log information, we should be able to pin down which trace spans are excessively large and limit those in our code.

This error can only occur with the Jaeger UDP transport, since OTLP uses gRPC which doesn't limit message size.

### DIFF
--- a/foundations/src/telemetry/tracing/output_jaeger_thrift_udp.rs
+++ b/foundations/src/telemetry/tracing/output_jaeger_thrift_udp.rs
@@ -60,8 +60,71 @@ fn get_reporter_bind_addr(settings: &JaegerThriftUdpOutputSettings) -> Bootstrap
 async fn do_export(reporter: JaegerCompactReporter, mut span_rx: SpanReceiver) {
     while let Some(span) = span_rx.recv().await {
         // NOTE: we are limited with a UDP dgram size here, so doing batching is risky.
-        if let Err(err) = reporter.report(&[span][..]).await {
+        let spans = [span];
+        if let Err(err) = reporter.report(&spans).await {
+            #[cfg(feature = "logging")]
+            if self::logging::is_msgsize_error(&err) {
+                self::logging::log_span_too_large_err(&err, &spans[0]);
+                continue;
+            }
+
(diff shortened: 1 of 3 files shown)


foundations/src/telemetry/tracing/output_jaeger_thrift_udp.rs
```diff
@@ -60,8 +61,65 @@ fn get_reporter_bind_addr(settings: &JaegerThriftUdpOutputSettings) -> Bootstrap
...
+
+    let top_tag = top_tag
+        .map(|t| format!(", top: {} @ {}", t.name(), tag_size(t)))
+        .unwrap_or_default();
+
+    crate::telemetry::log::error!(
+        "trace span exceeded thrift UDP message size limits";
+        "error" => %err,
+        "operation" => span.operation_name(),
+        "tags" => format!("count: {tag_count}, size: {tag_total}{top_tag}"),
```
inikulin, line 100: Let's be more explicit here that size is approximated

foundations/src/telemetry/tracing/output_jaeger_thrift_udp.rs
```diff
@@ -60,8 +61,65 @@ fn get_reporter_bind_addr(settings: &JaegerThriftUdpOutputSettings) -> Bootstrap
...
 }
+
+fn is_msgsize_error(err: &cf_rustracing::Error) -> bool {
+    err.concrete_cause::<io::Error>()
+        .and_then(io::Error::raw_os_error)
+        .map_or(false, |errno| errno == libc::EMSGSIZE)
+}
+
+#[cfg(feature = "logging")]
+fn span_too_large(err: &cf_rustracing::Error, span: &FinishedSpan) {
```
inikulin, line 84: Nit: let's rename to `log_span_too_large_err`

inikulin (approved):

inikulin (approved):

PR #96, opened 2024-11-18
Support configuring behaviour when excessive log nesting is reached

jgpaiva:
Previously, reaching too much log nesting caused foundations to panic. This allows disabling that behaviour and instead logging an error. It will only log when the threshold is passed, from then on it'll only ignore the nesting operation (i.e. the logger will remain unchanged).

### DIFF
--- a/foundations/src/telemetry/log/internal.rs
+++ b/foundations/src/telemetry/log/internal.rs
@@ -34,20 +34,39 @@ impl LoggerWithKvNestingTracking {
         }
     }
 
-    /// Increment the KV nesting level. You should call this every time you have replaced the
+    /// Increment the KV nesting level. You should call this before any time you're going to replace the
     /// logger with a child of itself.
     ///
-    /// If this returns true, you should first release any locks you may hold (they will be
-    /// poisoned) and then call the panic_from_too_much_nesting() associated function.
-    #[must_use]
-    pub(crate) fn has_too_much_nesting(&mut self) -> bool {
-        self.nesting_level += 1;
-        self.nesting_level > Self::MAX_NESTING
-    }
+    /// If this returns None, it will consume the logger lock, and you should not nest any further.
+    /// If panic_on_too_much_logger_nesting is enabled, instead of returning None this will free the
+    /// logger lock and then panic.
+    pub(crate) fn check_nesting_level(
+        mut current_log_lock: parking_lot::lock_api::RwLockWriteGuard<
+            parking_lot::RawRwLock,
+            LoggerWithKvNestingTracking,
+        >,
+    ) -> Option<
+        parking_lot::lock_api::RwLockWriteGuard<
+            parking_lot::RawRwLock,
+            LoggerWithKvNestingTracking,
+        >,
+    > {
+        current_log_lock.nesting_level = current_log_lock.nesting_level.saturating_add(1);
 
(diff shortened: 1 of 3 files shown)


foundations/src/telemetry/log/internal.rs
```diff
@@ -34,20 +35,39 @@ impl LoggerWithKvNestingTracking {
...
+            parking_lot::RawRwLock,
+            LoggerWithKvNestingTracking,
+        >,
+    ) -> Option<
+        parking_lot::lock_api::RwLockWriteGuard<
+            parking_lot::RawRwLock,
+            LoggerWithKvNestingTracking,
+        >,
+    > {
+        current_log_lock.nesting_level += 1;
```
bcottrell8738, line 55: We will eventually wrap around and start nesting again. I don't know how worried to be about this, but it might feel less icky if we incremented it only in the first and second arms of the match?

foundations/src/telemetry/log/internal.rs
```diff
@@ -1,4 +1,5 @@
 use super::init::LogHarness;
+use crate::telemetry::log::internal;
```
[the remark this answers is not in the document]
jgpaiva: ah, TY, this is bad copy-paste because I moved things around, originally the new code wasn't in this file

foundations/src/telemetry/log/internal.rs
jgpaiva: I was being a bit optimistic that we wouldn't try to nest that much. I'll fix it.

OmegaJak: Or `saturating_add`

bcottrell8738: saturating_add is good too!

jgpaiva: nice!

foundations/src/telemetry/log/internal.rs
```diff
@@ -34,20 +34,39 @@ impl LoggerWithKvNestingTracking {
...
+            parking_lot::RawRwLock,
+            LoggerWithKvNestingTracking,
+        >,
+    ) -> Option<
+        parking_lot::lock_api::RwLockWriteGuard<
+            parking_lot::RawRwLock,
+            LoggerWithKvNestingTracking,
+        >,
+    > {
+        current_log_lock.nesting_level.saturating_add(1);
```
bcottrell8738, line 54: We should take the result of the saturating_add, and assign it back to the nesting_level field.

jgpaiva: omg, TY for the callout. Fixed

foundations/src/telemetry/log/internal.rs
```diff
@@ -34,20 +34,39 @@ impl LoggerWithKvNestingTracking {
...
+            parking_lot::RawRwLock,
+            LoggerWithKvNestingTracking,
+        >,
+    ) -> Option<
+        parking_lot::lock_api::RwLockWriteGuard<
+            parking_lot::RawRwLock,
+            LoggerWithKvNestingTracking,
+        >,
+    > {
+        current_log_lock.nesting_level = current_log_lock.nesting_level.saturating_add(1);
```
nmldiegues, line 54: nicely caught

nmldiegues (approved):

PR #101, opened 2024-12-05
Add block buffered stdout for JSON output

aberglund-cf:


foundations/src/telemetry/log/init.rs
```diff
@@ -95,12 +97,29 @@ pub(crate) fn init(service_info: &ServiceInfo, settings: &LoggingSettings) -> Bo
...
             let buf = BufWriter::with_capacity(BUF_SIZE, File::create(file)?);
-            let drain = build_json_log_drain(buf);
+            let drain = build_json_log_drain(buf, false);
             AsyncDrain::new(drain).chan_size(CHANNEL_SIZE).build()
         }
+        (LogOutput::Stdout, log_fmt) => {
+            let stdout = unsafe { File::from_raw_fd(STDOUT_FILENO) };
+            let buf = BufWriter::with_capacity(BUF_SIZE, stdout);
+
+            match log_fmt {
```
aberglund-cf, line 107: I wanted to reduce the duplication a bit here by creating `stdout` in one place, but if the nested match is not the preferred style then I'm fine with keeping the pattern consistent with the other branches.

foundations/src/telemetry/log/init.rs
```diff
@@ -95,12 +96,29 @@ pub(crate) fn init(service_info: &ServiceInfo, settings: &LoggingSettings) -> Bo
         }
         (LogOutput::File(file), LogFormat::Json) => {
             let buf = BufWriter::with_capacity(BUF_SIZE, File::create(file)?);
-            let drain = build_json_log_drain(buf);
+            let drain = build_json_log_drain(buf, false);
             AsyncDrain::new(drain).chan_size(CHANNEL_SIZE).build()
         }
+        (LogOutput::Stdout, log_fmt) => {
```
inikulin, line 102: I think we don't need additional `LogOutput` type. `Terminal` is basically stdout

foundations/src/telemetry/log/init.rs
```diff
@@ -95,12 +96,29 @@ pub(crate) fn init(service_info: &ServiceInfo, settings: &LoggingSettings) -> Bo
...
+            let stdout = unsafe { File::from_raw_fd(STDOUT_FILENO) };
```
inikulin, line 103: Let's refactor this unsafe expression into a function, e.g. `unbuffered_stdout()` or something like that and add a code comment explaining why we need it this way

foundations/src/telemetry/log/init.rs
```diff
@@ -95,12 +96,29 @@ pub(crate) fn init(service_info: &ServiceInfo, settings: &LoggingSettings) -> Bo
...
+
+            match log_fmt {
+                LogFormat::Text => {
+                    let drain = TextDrain::new(PlainDecorator::new(buf))
+                        .build()
+                        .fuse();
+                    AsyncDrain::new(drain).chan_size(CHANNEL_SIZE).build()
+                },
+                LogFormat::Json => {
+                    let drain = build_json_log_drain(buf, false);
```
inikulin, line 114: We need to keep `flush` option enabled on JSON drain. It guarantees that JSON record will be flushed once finished even if underlying BufWriter is not full: https://docs.rs/slog-json/latest/slog_json/struct.JsonBuilder.html#method.set_flush

foundations/src/telemetry/log/init.rs
```diff
@@ -124,6 +125,14 @@ pub(crate) fn init(service_info: &ServiceInfo, settings: &LoggingSettings) -> Bo
     Ok(())
 }
 
+/// Opens fd 1 directly with a [`BUF_SIZE`] buffer.
+///
+/// [`io::Stdout`] is line buffered and adds unnecessary overhead for JSON output.
+fn block_buffered_stdout() -> BufWriter<File> {
```
inikulin, line 131: Technically line buffering is a case of block buffering, also "block buffering" is not very often used term in the Rust realm as far as I can tell and also having it at the start of the function name one might confuse it with a `verb` thinking that it blocks buffered stdout. So, maybe something slightly goofy, but more explicit would work better, like `stdout_writer_without_line_buffering`

inikulin (approved):

PR #105, opened 2025-01-08
Add RetryFileWriter to handle BrokenPipe errors

aberglund-cf:
These errors are recoverable if the pipe still exists and the reader may come back at some point.


foundations/src/telemetry/log/retry_writer.rs
```diff
@@ -0,0 +1,125 @@
...
+/// Gives up and return an error after the max number of attempts or if any other errors are encountered.
+///
+/// Intended for use when writing to a pipe (fifo) with a reader which may randomly close and then
+/// reopen read side.
+///
+/// *WARNING*: This may cause duplication of data as it's not possible to know what bytes
+/// where actually processed by the read side.
+pub(crate) struct RetryPipeWriter {
+    path: PathBuf,
+    pipe_file: Box<File>,
```
kornelski, line 17: `File` is a 4-byte integer, so you don't need to `Box` it.

kornelski (approved):

fisherdarling (approved):

foundations/src/telemetry/log/init.rs
```diff
@@ -88,14 +89,14 @@ pub(crate) fn init(service_info: &ServiceInfo, settings: &LoggingSettings) -> Bo
             let drain = build_json_log_drain(stdout_writer);
             AsyncDrain::new(drain).chan_size(CHANNEL_SIZE).build()
         }
-        (LogOutput::File(file), LogFormat::Text) => {
-            let drain = TextDrain::new(PlainDecorator::new(File::create(file)?))
-                .build()
-                .fuse();
+        (LogOutput::File(file_path), LogFormat::Text) => {
```
inikulin, line 92: I think the original issue was observed when we were writing to stdout (i.e. `Terminal`) and not to a file. iirc we really use file outputs only for testing/debugging purposes on dev machines

foundations/src/telemetry/log/retry_writer.rs
```diff
@@ -0,0 +1,125 @@
...
+use std::path::PathBuf;
+
+/// A writer which retries if an [`io::ErrorKind::BrokenPipe`] is returned.
+///
+/// Gives up and return an error after the max number of attempts or if any other errors are encountered.
+///
+/// Intended for use when writing to a pipe (fifo) with a reader which may randomly close and then
+/// reopen read side.
+///
+/// *WARNING*: This may cause duplication of data as it's not possible to know what bytes
```
inikulin, line 13: I think we should elaborate more here about what actually happens and why doing IO in this hot path is kind of safe. 

We do 10 retries here that would probably block the thread for a significant amount of time. Though, AsyncDrain helps us to not block the tokio worker threads as it spawns its own worker thread where it performs sync IO and this thread receives messages to write via a bound crossbeam channel.

If we block the channel for a long time the channel will start dropping messages. However, async drain is configured to emit error log entries in such a case with the number of dropped messages. Concern here is that we can drop them as well, basically having our writes busy looping a lot and us not having any observability into it.

I think, we should also have a counter of retries stored in the writer that reset once in a while (e.g. once in a minute), so we can check if we run into retrying in a small period of time and likely dropping lots of messages and bail in this case as well.

aberglund-cf: > Concern here is that we can drop them as well, basically having our writes busy looping a lot and us not having any observability into it.

I agree we shouldn't be dropping messages silently, I didn't think this code was dropping messages at all. My expectation (and intention) was that this loop would continue until a write succeeds or returns an error (causing a panic).

That leads me to the question, if dropped messages are reported then do we ever want this to panic on error on EPIPE? Is it preferable to prevent thrashing but otherwise simply continue to reconnect as long as the pipe exists?

inikulin: I think we should still be able to give up at some point to not busy loop if the output went away for some reason and won't come back. As for logging: we kind of will still log, but it will be unreliable. Anyway, it's better than nothing I think.

So, yeah, let's just keep it as is, just let's probably add a code comment summarising this discussion and that explains that these retries can actually lead to entries being dropped, but those drops will be reported as log entries (though, there's a chance that those reports will be dropped as well if retries occur again)

inikulin (approved):

PR #107, opened 2025-01-30
OXY-1514: only track sampled spans and through a setting track all spans.

fisherdarling:
We're seeing high lock contention when removing spans from the `LiveReferenceSet`. Currently,
we sample all spans that are created, even if those spans are unsampled. The purpose of this
is to capture graceful restart timeout issues which sometimes occur in our services. By fetching
the list of all spans in memory when a graceful restart hangs, we can figure out what is preventing
the restart from occuring within its timeout.

In this commit we only track _sampled_ spans. Most services which use foundations sample at a 1% rate,
meaning we effectively remove 99% of locks. This allows light debugability in all running services, at
any time you can get a list of traced spans. Although those spans should end up in jaeger/oltp at some
point[^1]. To make sure we can still determine the root cause of graceful restart issues, we add a setting
which forces all spans to be sampled (what happened previously).

[^1]: to be fair, sometimes some traces will remain in memory forever and never be dropped, so even
at 1% this can be useful.

### DIFF
--- a/foundations/src/telemetry/tracing/live/event_output.rs
+++ b/foundations/src/telemetry/tracing/live/event_output.rs
@@ -23,7 +23,9 @@ pub(crate) fn spans_to_trace_events(
 
     for span in spans {
         let span_ref = span.read();
-        let span_state = span_ref.context().unwrap().state();
+        let Some(span_state) = span_ref.context().map(|c| c.state()) else {
+            continue;
+        };
         let trace_id = span_state.trace_id().to_string();
         let name = span_ref.operation_name();
 


foundations/src/telemetry/settings/tracing.rs
```diff
@@ -26,6 +26,13 @@ pub struct TracingSettings {
 
     /// The strategy used to sample traces.
     pub sampling_strategy: SamplingStrategy,
+    /// Enable liveness tracking of all generated spans. Even if the spans are
+    /// unsampled. This can be useful for debugging potential hangs cause by
```
fisherdarling, line 30:
```suggestion
    /// unsampled. This can be useful for debugging potential hangs caused by
```

foundations/src/telemetry/settings/tracing.rs
```diff
@@ -26,6 +26,13 @@ pub struct TracingSettings {
...
+    /// some objects remaining in memory.  The default value is false, meaning
```
fisherdarling, line 31:
```suggestion
    /// some objects remaining in memory.  The default value is `false`, meaning
```

foundations/src/telemetry/tracing/live/mod.rs
```diff
@@ -1,16 +1,20 @@
...
+use parking_lot::RwLock;
 use std::sync::Arc;
 use std::time::SystemTime;
 
-type SharedSpanInner = Arc<parking_lot::RwLock<Span>>;
-pub(crate) type SharedSpanHandle = Arc<LiveReferenceHandle<SharedSpanInner>>;
+pub(crate) use live_reference_set::LiveReferenceHandle;
+
+use crate::telemetry::tracing::internal::SharedSpanHandle;
+// pub(crate) type SharedSpanHandle = Arc<SharedSpanHandle>;
```
fisherdarling, line 13:
```suggestion
```

foundations/src/telemetry/tracing/internal.rs
```diff
@@ -15,37 +17,49 @@ pub(crate) type Tracer = cf_rustracing::Tracer<BoxSampler<SpanContextState>, Spa
 /// Shared span with mutability and additional reference tracking for
 /// ad-hoc inspection.
 #[derive(Clone, Debug)]
-pub(crate) struct SharedSpanInner(SharedSpanHandle);
+pub(crate) enum SharedSpanHandle {
+    Tracked(Arc<LiveReferenceHandle<Arc<RwLock<Span>>>>),
+    Unsampled(Arc<RwLock<Span>>),
```
fisherdarling, line 22: bikeshed?
```suggestion
    Untracked(Arc<RwLock<Span>>),
```

foundations/src/telemetry/settings/tracing.rs
```diff
@@ -26,6 +26,13 @@ pub struct TracingSettings {
...
+    /// only sampled spans are tracked.
+    ///
+    /// To get a json dump of the currently active spans, query: `/debug/traces`
+    pub track_all_spans: bool,
```
inikulin, line 35: I think we need to make the whole liveness tracking functionality, i.e. introduce `LivenessTrackingSettings` with `enabled` field and also put this field in there too.

foundations/src/telemetry/settings/tracing.rs
```diff
@@ -26,6 +26,13 @@ pub struct TracingSettings {
 
     /// The strategy used to sample traces.
     pub sampling_strategy: SamplingStrategy,
+    /// Enable liveness tracking of all generated spans. Even if the spans are
+    /// unsampled. This can be useful for debugging potential hangs cause by
+    /// some objects remaining in memory.  The default value is false, meaning
+    /// only sampled spans are tracked.
+    ///
+    /// To get a json dump of the currently active spans, query: `/debug/traces`
```
inikulin, line 34: Nit: worth mentioning here that we are talking about querying telemetry server's route here. It's confusing out of the context

foundations/src/telemetry/tracing/live/mod.rs
```diff
@@ -1,34 +1,108 @@
...
+use parking_lot::RwLock;
 use std::sync::Arc;
 use std::time::SystemTime;
 
-type SharedSpanInner = Arc<parking_lot::RwLock<Span>>;
-pub(crate) type SharedSpanHandle = Arc<LiveReferenceHandle<SharedSpanInner>>;
+pub(crate) use live_reference_set::LiveReferenceHandle;
+
+use crate::telemetry::settings::LivenessTrackingSettings;
+use crate::telemetry::tracing::internal::SharedSpanHandle;
```
inikulin, line 13: Nit: here and in other places: don't separate those `use`'s from the rest of the non-public imports

inikulin (approved):

bobrik:
Can we get a release with this as well?

PR #116, opened 2025-03-03
Re-export Hyper types used in public API

tsheinen:
Foundations uses Hyper types in it's public API for `TelemetryServerRoute` which means any users will need to add an explicit dependency on the same version of hyper and ensure it stays matching. Re-exporting the types will allow users to explicitly use the same types as foundations is using.

### DIFF
--- a/examples/http_server/main.rs
+++ b/examples/http_server/main.rs
@@ -15,12 +15,15 @@ use self::settings::{EndpointSettings, HttpServerSettings, ResponseSettings};
 use anyhow::anyhow;
 use foundations::cli::{Arg, ArgAction, Cli};
 use foundations::settings::collections::Map;
-use foundations::telemetry::{self, log, tracing, TelemetryConfig, TelemetryContext};
+use foundations::telemetry::{
+    self,
+    hyper::{Body, Request, Response},
+    log, tracing, TelemetryConfig, TelemetryContext,
+};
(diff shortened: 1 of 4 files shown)


TheJokr:
Rebased and altered the PR slightly. The hyper 1.x bump meant that body handling got quite a bit more complicated, so we would need to export more types. Instead, I chose to re-export all of `hyper` under a new `telemetry::reexports` module and introduce an alias for the `BoxBody` type from `http-body-util`.

foundations/src/telemetry/server/router.rs
```diff
@@ -16,12 +16,12 @@ use std::collections::HashMap;
...
 
+/// Body type used in [`TelemetryServerRoute`] responses.
+pub type TelemetryRouteBody = BoxBody<Bytes, crate::Error>;
+
 /// Future returned by [`TelemetryServerRoute::handler`].
 pub type TelemetryRouteHandlerFuture =
-    BoxFuture<'static, std::result::Result<Response<BoxBody<Bytes, BoxError>>, Infallible>>;
-
-/// Error type returned by [`TelemetryRouteHandlerFuture`].
-pub type BoxError = Box<dyn std::error::Error + Send + Sync>;
```
TheJokr, line 24: This alias was a duplicate of `crate::Error`

PR #118, opened 2025-03-10
feat: Add tracing-rs-compat feature with a tracing-rs compatible logging drain

lpraneis:
Services and libraries that use `foundations` for logging must configure how the logs are output. This works well for services that want to forward logs to stdout / stderr for systemd but doesn't work as well for projects using libraries that use this `foundations` logging  as they may have their own complex logging machinery in place.

This PR adds a new feature, `tracing-rs-compat`, that enables the option to use [`tracing-slog`](https://crates.io/crates/tracing-slog) to install a `slog::Drain` that will forward any logs to the currently installed tracing subscriber. This makes it possible for any projects heavily invested in the [`tracing`](https://crates.io/crates/tracing) ecosystem to at least ingest the logs emitted using the logging macros in `foundations`.

Open to any feedback on the design here!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,8 @@ tokio = "1.41.0"
 thread_local = "1.1"
 tikv-jemallocator = "0.5"
 tikv-jemalloc-ctl = "0.5"
+tracing-slog = "0.3.0"
+tracing-subscriber = "0.3"
 yaml-merge-keys = "0.5"
 
 # needed for minver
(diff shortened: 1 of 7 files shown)


inikulin (approved):

lpraneis:
@inikulin rebased on main with the clippy fixes, can you please approve CI re-run?

TheJokr (approved):

PR #119, opened 2025-04-30
Update deps with minimal changes

lorlouis:
I updated dependencies which required a minimal amount of work to upgrade.

Some of the larger libraries, `serde_yaml`,  `hyper`, and the Prometheus crates, would require significant code churn to get up to date, so I left them alone for now.

I also updated serde deny's config to v2 format since v1 got dropped a while back.

v2 migration context
https://github.com/EmbarkStudios/cargo-deny/pull/611

I also fixed a few clippy lints/deprecation warnings.


deny.toml
```diff
@@ -24,8 +23,33 @@ unknown-git = "allow"
...
 
 [licenses]
-unlicensed = "allow"
-allow-osi-fsf-free = "either"
-copyleft = "deny"
-allow = ["MPL-2.0"]
+private.ignore = true
+unused-allowed-license = "allow"
+# List of explicitly allowed licenses
+# These are explicitly allowed by Cloudflare.
```
cbranch, line 29: foundations is a general-purpose library that is also used outside Cloudflare and these licenses are not universally suitable. Can you revise the allowlist to match the original intent?

deny.toml
```diff
@@ -1,15 +1,14 @@
...
-unsound = "deny"
-severity-threshold = "medium"
+unmaintained = "all"
 ignore = [
-    # No fix available, but only applies to Windows, hence not a problem for us.
-    "RUSTSEC-2021-0145",
+    # NOTE: paste which is unmaintained, but
+    # is a dependency of tykv-jemalloc-sys
+    "RUSTSEC-2024-0436",
+    # NOTE: serde_yaml will have to be replaced with
```
cbranch, line 7: Can you remove the note? This change isn't necessarily needed or desired.

deny.toml
lorlouis: I've made the switch and opted to only list the licences currently in use.
Cargo deny will let us know if a licence is osi or fsf-free so we can add new ones as needed.

cbranch (approved):

lorlouis:
It looks like the minimal-versions CI step can't be made to work without upgrading a few significant libraries.
It currently fails on some transitive dependencies which are expecting an older version of the compiler and explicitly declare features which have been stabilised.

Since it won't break this version of the library any more than it currently is, I would suggest merging it anyway, and I'll try to update hyper, tonic and a few other big crates in another patch.

fisherdarling:
Thanks for the initial work! This is has been superseded by #132

PR #120, opened 2025-05-02
Remove default constraint from `Settings` trait

zakcutner:
This required that fields of structs that derive `Settings` implement `Default`, even if a default implementation of the field was provided.

This occurred in particular when trying to use a `NonZero` type within a settings struct:

```rust
pub struct MySettings {
    #[serde(default = "MySettings::default_my_field")]
    pub my_field: NonZeroUsize,
}

impl MySettings {
    fn default_my_field() -> NonZeroUsize {
        return NonZeroUsize::new(1).unwrap()
    }
}
```

Instead, this change moves the `Default` constraint to only where the `Settings` is used with a `Cli`.

### DIFF
--- a/foundations/src/cli.rs
+++ b/foundations/src/cli.rs
@@ -37,7 +37,7 @@ pub struct Cli<S: Settings> {
     pub arg_matches: ArgMatches,
 }
 
-impl<S: Settings> Cli<S> {
+impl<S: Default + Settings> Cli<S> {
     /// Bootstraps a new command line interface (CLI) for the service.
     ///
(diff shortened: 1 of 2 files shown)


vcfxb (approved):
LGTM!

inikulin:
I think it contradicts some of the paradigms established here: https://docs.rs/foundations/latest/foundations/settings/index.html

For cases like yours the suggestion is to implement a wrapper type that implement `Default`. Likely in your case it also will give some semantic meaning to your type (e.g. `WorkerCount`)

PR #121, opened 2025-05-08
ZTC-2169: Allow use of memory-profiling feature with no other features

OmegaJak:
For our service, we only want to use the `memory-profiling` feature, but not the other telemetry-related features. However, currently, the `telemetry` module is not compiled if `memory-profiling` is enabled but none of the others are.

This updates the conditional compilation so that, using only the `memory-profiling` feature, the `MemoryProfiler` can be accessed.

### DIFF
--- a/foundations/Cargo.toml
+++ b/foundations/Cargo.toml
@@ -145,6 +145,7 @@ memory-profiling = [
     "dep:tikv-jemalloc-ctl",
     "dep:tempfile",
     "dep:tokio",
+    "dep:serde",
     "jemalloc",
 ]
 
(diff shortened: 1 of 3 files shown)


cbranch (approved):

TheJokr (approved):

OmegaJak:
CI is failing due to lints in new Rust versions. Addressed in https://github.com/cloudflare/foundations/pull/122, which should be merged first

OmegaJak:
Rebased onto latest w/ clippy fixes from #122, CI should pass now if someone can trigger it

PR #126, opened 2025-07-17
Add API for automatically recording trace spans duration to a histogram

evanrittenhouse:
This adds an API for automatically tracking span durations with a `TimeHistogram`. The durations are stored regardless of if the underlying span is sampled.


oga35767-eng (approved):

oga35767-eng:
👍

PR #127, opened 2025-07-17
Add compile-time validation for histogram buckets in foundations-macros

LloydW93:
I recently had a bug in a service using foundations where the buckets in my histogram were not in ascending order because of a typo. Instead of this being detected and handled in any way, I just had a broken metric.

This change adds compile time checks when using the metrics proc macro to, if the buckets are provided directly, check the ordering. It doesn't check if they're passed from a variable or function or any other method, and I think the only way to do that would be runtime validation - the binary would consistently fail to start immediately, but have kept this constrained to a compile-time check for now to get thoughts/feedback.

### DIFF
--- a/foundations-macros/src/metrics/mod.rs
+++ b/foundations-macros/src/metrics/mod.rs
@@ -9,6 +9,7 @@ use syn::{
 };
 
 mod parsing;
+mod validation;
 
 #[derive(FromMeta)]
 struct MacroArgs {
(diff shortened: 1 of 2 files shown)


MichaHoffmann (approved):

TheJokr (approved):

fisherdarling:
rebased

fisherdarling:
fixed some clippy issues

PR #128, opened 2025-08-13
Add unix domain socket for telemetry server

gpgabriel:


foundations/src/telemetry/driver.rs
```diff
@@ -49,8 +48,8 @@ impl TelemetryDriver {
     ///
     /// Returns `None` if the server wasn't spawned.
     #[cfg(feature = "telemetry-server")]
-    pub fn server_addr(&self) -> Option<SocketAddr> {
-        self.server_addr
+    pub fn server_addr(&self) -> Option<String> {
```
nox, line 51: Why is this not a `TelemetryServerAddr`?

foundations/src/telemetry/server/mod.rs
```diff
@@ -14,18 +14,33 @@ use std::net::SocketAddr;
...
 use tokio::net::TcpListener;
+#[cfg(unix)]
+use tokio::net::{TcpStream, UnixListener, UnixStream};
 use tokio::sync::watch;
 
 mod router;
 
 use router::Router;
+
+trait AsyncReadWrite: AsyncRead + AsyncWrite + Send + Unpin + 'static {}
```
nox, line 27: Why is this needed?

nox (approved):
The changes seem reasonable to me but note that this includes breaking changes in public API so that requires a major foundations release.

foundations/src/telemetry/driver.rs
gpgabriel: `server_addr` is only used for logging, always needed as a String

foundations/src/telemetry/server/mod.rs
gpgabriel: I added this to be able to store each kind of listener in the same variable - https://github.com/cloudflare/foundations/pull/128/files#diff-caee81f82e8cb546f9fde63dc42376101dc4498dfcdf5b7af0ddd8890312338fR135-R137

foundations/src/lib.rs
```diff
@@ -69,7 +69,7 @@ mod utils;
 #[cfg(feature = "cli")]
 pub mod cli;
 
-#[cfg(feature = "settings")]
+#[cfg(any(feature = "settings", feature = "telemetry-server"))]
```
nox, line 72: Let's put `ListenAddr` elsewhere, rather than expose this module when `settings` is off.

foundations/src/settings/addr.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[cfg(unix)]
+impl From<std::path::PathBuf> for ListenAddr {
+    fn from(path: std::path::PathBuf) -> Self {
+        ListenAddr::Unix(path)
+    }
+}
+
+#[cfg(unix)]
+impl From<&str> for ListenAddr {
```
nox, line 49: I feel like these impls can be misused and people could end up doing `ListenAddr::from("127.0.0.1:0")`.

foundations/src/telemetry/settings/server.rs
```diff
@@ -16,15 +13,15 @@ pub struct TelemetryServerSettings {
 
     /// Telemetry server address.
     #[serde(default = "TelemetryServerSettings::default_addr")]
-    pub addr: SocketAddr,
+    pub addr: ListenAddr,
```
nox, line 16: This is still a breaking change, but we should release foundations 5 anyway for the hyper 1 switch.

nox (changes requested):

foundations/src/settings/addr.rs
gpgabriel: Fully removed them for now

foundations/src/telemetry/driver.rs
```diff
@@ -47,10 +48,23 @@ impl TelemetryDriver {
...
+        match self.addr.as_ref()? {
+            ListenAddr::Tcp(addr) => Some(*addr),
+            #[cfg(unix)]
+            ListenAddr::Unix(_) => None,
+        }
+    }
+
+    /// Address of the temeletry server.
+    ///
+    /// Returns `None` if the server weasn't spawned.
```
marmeladema, line 64: typo: wasn't

nox:
You need to rebase on top of https://github.com/cloudflare/foundations/pull/130 to fix the clippy issues.

nox:
I rebased it.

foundations/src/telemetry/driver.rs
```diff
@@ -45,12 +46,12 @@ impl TelemetryDriver {
         }
     }
 
-    /// Address of the telemetry server.
+    /// Address of the temeletry server.
```
marmeladema, line 49: Typo: temeletry -> telemetry

PR #129, opened 2025-08-20
Add tracing::w3c_traceparent() helper

bobrik:


foundations/src/telemetry/tracing/mod.rs
```diff
@@ -294,6 +294,23 @@ pub fn state_for_trace_stitching() -> Option<SerializableTraceState> {
         .map(|c| c.state().clone())
 }
 
+/// Returns the value to be used as a W3C traceparent header.
+///
+/// See: https://www.w3.org/TR/trace-context/#traceparent-header
+///
+/// Returns `None` if the current span is not sampled and don't have associated trace.
```
mattrighetti, line 301: nit

```suggestion
/// Returns `None` if the current span is not sampled and doesn't have an associated trace.
```

bobrik: I copy-pasted this one from above, so I fixed both.

PR #132, opened 2025-08-27
Update deps with minimal changes

fisherdarling:
Minimal dependency updates.

Based on this PR: https://github.com/cloudflare/foundations/pull/119

- [x] need to add attribution
- [x] go through the rest of the deps
  - updated darling -> 0.21

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,42 +27,42 @@ debug = 1
 anyhow = "1.0.75"
 foundations = { version = "4.5.0", path = "./foundations" }
 foundations-macros = { version = "4.5.0", path = "./foundations-macros", default-features = false }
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -27,63 +27,64 @@ debug = 1
...
 slog-json = "2.3"
 slog-term = "2.4"
 tempfile = "3.7"
 tokio = "1.41.0"
 thread_local = "1.1"
-tikv-jemallocator = "0.5"
-tikv-jemalloc-ctl = "0.5"
+tikv-jemallocator = "0.6"
+tikv-jemalloc-ctl = "0.6"
+tower = "0.5.2"
```
TheJokr, line 75: this should go below in the `needed for minver` list, we don't use tower directly anywhere

PR #156, opened 2025-11-25
Add #[with_removal] attribute to #[metrics] macro

TheJokr:
`#[with_removal]` generates two additional functions that allow to
remove individual label sets from a metric family as well as clear the
family completely. This is useful when metrics are used for dynamic
data, such as peer RTT in a P2P system.

The feature is hidden behind `cfg(foundations_unstable)`, which allows
us to iterate on the API in the future. Both proc-macro unit tests as
well as an integration test are included in the change.

Tests on this PR will fail until nox/prometools#6 is merged and released, which
exposes the necessary `Family` APIs to implement this feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,18 @@ publish = false
 [profile.release]
 debug = 1
 
+[workspace.lints.rust.unexpected_cfgs]
+level = "warn"
+check-cfg = [
+    'cfg(foundations_generic_telemetry_wrapper)',
+    'cfg(foundations_unstable)',
+    'cfg(tokio_unstable)',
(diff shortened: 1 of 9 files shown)


fisherdarling (approved):
Looks good! 

General q, is it possible to have an API of:
```
regular::dynamic("foo").remove();
```
Rather than having the `_remove` suffix?

TheJokr:
> General q, is it possible to have an API of:
> 
> ```
> regular::dynamic("foo").remove();
> ```

Yes, good idea! I'll make some changes to the PR.

oga35767-eng (approved):

PR #157, opened 2025-11-27
Improve performance of inactive tracing spans

TheJokr:
### Improve performance of log/tracing harness access

- Switch to `std::sync::{OnceLock, LazyLock}` for better storage
   efficiency and std's futex-based backend. These are available since
   Rust 1.80.
- Avoid unnecessary LazyLock initialization of `NOOP_HARNESS` if a
   proper harness is configured. Previously, `.unwrap_or(&NOOP_HARNESS)`
   would cause the Deref impl to be called even if the reference was
   never used. (Due to Deref's side effects.)
- Wrap harnesses in crossbeam's `CachePadded` to avoid false sharing
   with other global variables. This is important to ensure OnceLock's
   initialization check is not slowed down by dirty cache lines.

### Remove globally-shared INACTIVE_SPAN RwLock

This RwLock was a heavy contention point in applications that generate a
lot of spans, as every span operation must lock and unlock the RwLock.
Each lock/unlock is an atomic RMW and thus dirties the cache line for
all other cores.

We can avoid the global singleton `INACTIVE_SPAN` entirely with some
changes to our internal `SharedSpanHandle` API.

### Remove internal once_cell usages

There are two spots left where we can't remove once-cell yet:
- In `memory_profiler.rs`, we use `OnceCell::get_or_try_init`. This
   function is not stabilized in std yet.
- The `foundations::security::allow_list` macro generates OnceCell
   instances in user code. Changing this requires a major bump.

### DIFF
--- a/foundations/src/telemetry/tracing/init.rs
+++ b/foundations/src/telemetry/tracing/init.rs
@@ -3,35 +3,38 @@ use super::output_jaeger_thrift_udp;
 use crate::telemetry::scope::ScopeStack;
 use crate::telemetry::settings::{SamplingStrategy, TracesOutput, TracingSettings};
 use crate::telemetry::tracing::live::ActiveRoots;
+use crate::telemetry::tracing::rate_limit::RateLimitingProbabilisticSampler;
 use crate::{BootstrapResult, ServiceInfo};
+use cf_rustracing::sampler::{NullSampler, PassiveSampler, Sampler};
 use cf_rustracing_jaeger::span::SpanReceiver;
+use crossbeam_utils::CachePadded;
 use futures_util::future::BoxFuture;
-use once_cell::sync::{Lazy, OnceCell};
+use std::sync::{LazyLock, OnceLock};
 
 #[cfg(feature = "telemetry-otlp-grpc")]
 use super::output_otlp_grpc;
 
-use cf_rustracing::sampler::{PassiveSampler, Sampler};
 #[cfg(feature = "testing")]
 use std::borrow::Cow;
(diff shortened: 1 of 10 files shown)


foundations/src/telemetry/tracing/init.rs
```diff
@@ -3,35 +3,38 @@ use super::output_jaeger_thrift_udp;
...
 
-use crate::telemetry::tracing::rate_limit::RateLimitingProbabilisticSampler;
-
-static HARNESS: OnceCell<TracingHarness> = OnceCell::new();
+// These singletons are accessed _very often_, and each access requires an atomic load to
+// ensure initialization. Make sure nobody else invalidates our cache lines.
+static HARNESS: CachePadded<OnceLock<TracingHarness>> = CachePadded::new(OnceLock::new());
 
-static NOOP_HARNESS: Lazy<TracingHarness> = Lazy::new(|| {
-    let (noop_tracer, _) = Tracer::new(RateLimitingProbabilisticSampler::default().boxed());
```
TheJokr: `RateLimitingProbabilisticSampler::default()` uses a sampling rate of `0.0`, which is equivalent to just using NullSampler directly.

fisherdarling (approved):

oga35767-eng (approved):

foundations/src/telemetry/tracing/init.rs
oga35767-eng: 👍

oga35767-eng (approved):

PR #160, opened 2025-12-17
feat: panic and sentry hook for foundations

fisherdarling:
This PR adds a panic and sentry hook to foundations. The panic hook will increment a metric and log the fact a panic occurred, where the sentry hook only increments a metric. The panic hook uses foundations' logging functionality _if_ it's initialized, otherwise, it will print a JSON formatted log to stderr.

The two metrics emitted by foundations are: `panics_total` and `sentry_events_total{label=<...>}`.

To easily test foundations' global metrics, I migrated CI to use `cargo-nextest`. Also, I moved the target build testing to its own job (just to separate build/test). Finally, I generated a simple `AGENTS.md` which we can start refining to help with future PRs.

### DIFF
--- a/foundations/Cargo.toml
+++ b/foundations/Cargo.toml
@@ -45,7 +45,7 @@ platform-common-default = [
 ]
 
 # Sentry integration for fatal error tracking
-sentry = ["dep:sentry-core", "dep:serde"]
+sentry = ["dep:sentry-core", "dep:serde", "metrics"]
 
 # A subset of features that can be used both on server and client sides. Useful for libraries
 # that can be used either way.
--- a/foundations/src/panic/hook.rs
+++ b/foundations/src/panic/hook.rs
@@ -1,9 +1,6 @@
 //! Panic hook implementation for tracking panics.
 
 use std::panic::{self, PanicHookInfo};
-use std::sync::OnceLock;
-
-pub(crate) static HOOK_INSTALLED: OnceLock<()> = OnceLock::new();
 
 /// Install the panic hook.
 ///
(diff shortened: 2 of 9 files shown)


foundations/src/alerts/panic.rs
```diff
@@ -0,0 +1,133 @@
...
+        return;
+    }
+
+    // Fallback to printing structured JSON to stderr
+    let location_str = location
+        .map(|l| format!("{}:{}:{}", l.file(), l.line(), l.column()))
+        .unwrap_or_else(|| "<unknown>".to_string());
+
+    // Use serde_json for proper JSON escaping of all control characters
+    let json_output = serde_json::json!({
```
fisherdarling: Used for escaping the input `payload`. A little heavy, but most `foundations` users will likely have `serde_json` in their dep tree.

foundations/Cargo.toml
```diff
@@ -209,6 +213,7 @@ prometheus-client = { workspace = true, optional = true }
 prometools = { workspace = true, optional = true, features = ["serde"] }
 rand = { workspace = true, optional = true }
 serde = { workspace = true, optional = true, features = ["derive", "rc"] }
```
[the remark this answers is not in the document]
fisherdarling: Put `serde_json` in as optional, forgot to mark it as so!

foundations/src/panic/hook.rs
```diff
@@ -0,0 +1,72 @@
...
+pub(crate) static HOOK_INSTALLED: OnceLock<()> = OnceLock::new();
+
+/// Install the panic hook.
+///
+/// Returns `true` if this is the first installation, `false` if the hook
+/// was already installed (subsequent calls are no-ops).
+///
+/// See the module-level docs for more information: [`crate::panic`]
+pub fn install_hook() -> bool {
+    let first_install = HOOK_INSTALLED.set(()).is_ok();
```
TheJokr: same here with the AtomicBool. Or if you 100% want to make sure this is not called anywhere in parallel, use `get_or_init` and install the hook inside the init function.

foundations/src/sentry/metrics.rs
```diff
@@ -0,0 +1,11 @@
+//! Sentry event related metrics.
+
+use super::Level;
+use crate::telemetry::metrics::Counter;
+
+/// Sentry metrics.
+#[crate::telemetry::metrics::metrics(crate_path = "crate", unprefixed)]
+pub mod sentry_events {
```
TheJokr: what do you think about calling this `mod sentry` instead and have the metric be `events_total`?

that makes the names for other metrics in the future more sensible, e.g. `sentry_requests_failed` for the POST requests to the sentry API.

fisherdarling: Fixed! Makes a lot of sense to me :)

PR #166, opened 2026-01-28
Re-export http-body-util under telemetry reexports

tsheinen:
TelemetryRouteBody aliases `BoxBody` from `http-body-util`. However, to construct `BoxBody` we still need something which implements `Body`. The tests use `Full` from `http-body-util`.

follows from https://github.com/cloudflare/foundations/pull/116. I'm open to the idea that perhaps we should also just alias `Full` or add a constructor to `TelemetryRouteBody` but i erred towards exporting the entire crate given that `hyper` is already exported.


TheJokr (approved):

oga35767-eng (approved):

oga35767-eng:
👍

PR #168, opened 2026-02-12
feat: allow manually shutting down logging from TelemetryDriver

### LINKED ISSUE #114: Logs may be discarded if the program exits before AsyncDrain can write
`slog-async` drains will attempt to flush and join the writer thread on [drop](https://github.com/slog-rs/async/blob/master/lib.rs#L551). Foundations holds a reference to the async drain in a static OnceCell so it's never dropped. This means that if a program logs and then exits the writer thread will not be joined and that log might not be written. 

I'm thinking the least surprising solution is to just hook atexit and make sure it gets dropped with some changes to `HARNESS` to have sufficient interior mutability. 

`slog-async` also offers [AsyncGuard](https://github.com/slog-rs/async/blob/master/lib.rs#L551) which when dropped will tell the async thread to finish and then join on it. That could be added to `TelemetryDriver` or something to make sure the logs get flushed when it goes out of scope. IMO given currently logger semantics (stored globally, may only be called once, etc) it's simpler to just flush it on exit.

AkshatM:
Fixes #114, which describes the bug in detail. slog provides an `AsyncGuard` implementation which is designed to ensure log flushes on drops even if the `Async` logger is captured in a global variable.

To ensure backwards compatibility, we now return an AsyncGuard that can be manually dropped, allowing existing users to continue to rely on current semantics but guarantee logging is shut down when desired. Credit to @TheJokr for the suggestion. 

The most complicated part of this feature is its interaction with the `logging` feature. Since it only makes sense if `logging` is enabled, I have conditionally compiled accordingly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,12 +97,10 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: "recursive"
-      - name: Install Rust
-        uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
       - name: Run cargo deny checks
         uses: EmbarkStudios/cargo-deny-action@v2
         with:
+          rust-version: stable
           command: check
 
   clippy:


AkshatM:
Received confirmation from our internal user that this fix resolves the log flush behaviour as expected by calling `driver.shutdown_logger()` as described before program exit.

PR #198, opened 2026-04-16
fix: Default back to serde_yaml, with serde-saphyr opt-in

TheJokr:
serde-saphyr is more strict in terms of YAML spec compliance than
serde_yaml, which may cause services to break if they are using invalid
YAML in their configuration files that was accepted by serde_yaml.

We keep serde-saphyr as an opt-in feature for services that do not rely
on invalid YAML and which want to avoid the unmaintained serde_yaml
crate.

Includes release commit for `foundations 5.6.3`. This is not a breaking change,
since the YAML parser is purely an implementation detail.


TheJokr:
FYI the new files under `tests/data/serde-yaml` are exactly the same as we had under `tests/data` prior to #181. You can verify with `git diff 576dddc958c25169a8d8d2d2598ae3ba71bc2c82~..origin/lblocher/saphyr-opt-in -- foundations/tests/data` locally.

fisherdarling (approved):

PR #201, opened 2026-04-23
Bugfix release 5.6.5

TheJokr:
### fix: Revert initialization of metrics in telemetry testing

This partially reverts #199. The attempted fix in that commit uses `service_info!()` to initialize the metrics registries in `TestTelemetryContext::new()`. Cargo sets environment variables separately for each crate being compiled, so `service_info!()` resolved `name_in_metrics` to "foundations". What we wanted is to use the name of the crate being _tested_.

That wouldn't necessarily be correct either, as a crate could set a completely custom `name_in_metrics` in its `foundations::telemetry::init` call. Instead, we now make sure to not initialize the metrics registries altogether. I added a test to enforce this going forward.

###  fix: Use `num_tasks` setting for Jaeger trace reporter tasks

This was accidentally set to `max_batch_size`, and linters didn't notice the unused `num_tasks` setting because the field is `pub`.

### DIFF
--- a/foundations/src/telemetry/tracing/init.rs
+++ b/foundations/src/telemetry/tracing/init.rs
@@ -77,16 +77,10 @@ pub(super) fn create_tracer_and_span_rx(
     };
 
     if let Some(cap) = settings.max_queue_size {
-        #[cfg(feature = "metrics")]
-        super::metrics::tracing::max_queue_size().set(cap.get() as u64);
-
         let (consumer, span_rx) = super::channel::channel(cap);
         let tracer = Tracer::with_consumer(sampler, consumer);
         Ok((tracer, span_rx))
     } else {
-        #[cfg(feature = "metrics")]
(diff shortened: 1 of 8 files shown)


foundations/src/telemetry/tracing/init.rs
```diff
@@ -122,6 +116,15 @@ pub(crate) fn init(
     // Only spawn the futures if we are actually initializing the harness
     let mut res = Ok(None);
     HARNESS.get_or_init(|| {
+        #[cfg(feature = "metrics")]
+        {
+            let max_queue_size = settings
+                .max_queue_size
+                .map(std::num::NonZeroUsize::get)
+                .unwrap_or(usize::MAX);
+            super::metrics::tracing::max_queue_size().set(max_queue_size as u64);
```
TheJokr, line 125: we move the max_queue_size().set(...) call here so it 1) only runs in proper init (not testing) and 2) only if there isn't already a global tracer

fisherdarling (approved):

Issue #158, opened 2025-12-01
Breaking changes to consider for foundations 6


TheJokr:
This is a tracking issue to compile breaking changes we want to consider before releasing foundations 6.

TheJokr:
Metric functions with and without labels have a subtle difference in behavior: unlabeled metrics are emitted with their default value immediately, while labeled metrics are only created on first usage. (Naturally, since we can't know the labels ahead of time.)

This can cause a lot of uninteresting metrics to be emitted when only part of a metrics-enabled library is actually executed.

Making unlabeled metrics behave like labeled ones, by initializing them on first use, would make foundations' metrics handling more consistent and avoid wasting storage in such cases. However, zero values in prometheus are useful (e.g. for `rate()` calculations), so maybe we should offer both options?

TheJokr:
~~We should upgrade sentry to the latest major version. Since we expose a sentry before_send hook, this needs to happen in a major release.~~

E: We decided to extract the sentry functionality into a separate crate instead. This crate can make major releases more often. #203

TheJokr:
Our current metrics implementation is tightly coupled to a specific (pre-release) `prometheus-client` version. This has caused us some hassle in the recent past already (see #213). Instead, it would be nice if we had a dedicated `foundations-metrics` crate which contains our own global singleton registry. This can (hopefully) be stable across many foundations major versions, such that they can coexist in the same binary.

`foundations-metrics` should have a minimal API:
- A trait that metrics implement to encode into an output format
- A function to register a metric in the global registry (either a single `Box<dyn OurMetricTrait>` or a vec of boxed metrics)
- A function to collect all registered metrics for encoding

(Writing this down now, it seems very similar to the API of the non-official `prometheus` crate: https://docs.rs/prometheus. Maybe we just switch to that as our backing store?)

With this breaking change we also want to support protobuf metrics output (required for native histograms), so the output format should be protobuf. We can convert from protobuf to OpenMetrics text on the fly.

TheJokr:
Similar to the above, we should make it easy for libraries to use foundations' logging and tracing features across major versions. This requires splitting the APIs for those into separate crates (e.g., `foundations-logging`, `foundations-tracing`), or explicitly documenting external facades that can be used (`slog`, `cf-rustracing`).

As with metrics, the API for these crates needs to be very stable, so should be kept limited. For example, output handling should stay in `foundations`.

TheJokr:
[...]

PR #202, opened 2026-04-24
Update dependencies.

orium:


foundations/Cargo.toml
```diff
@@ -258,7 +258,6 @@ zeroize = { workspace = true, optional = true }
 # needed for minver purposes
 ahash = { workspace = true, optional = true }
 annotate-snippets = { workspace = true, optional = true }
-async-stream = { workspace = true, optional = true }
```
orium, line 261: Not used anymore. Technically a breaking change if someone was setting the `async-stream` feature.

Cargo.toml
```diff
@@ -107,14 +108,12 @@ zeroize = "1.8"
...
 annotate-snippets = "0.12.13"
-async-stream = "0.3"
 encoding_rs_io = "0.1.2"
-local-ip-address = "0.6.5"
+local-ip-address = "0.6.6"
 lock_api = "0.4.7"
 log = "0.4.27"
-neli = "0.6.5"
-neli-proc-macros = "0.1.4"
-nix = "0.30"
```
orium, line 117: moved to the section above: it's only used in tests (and there's no transitive use)

foundations/src/settings/mod.rs
```diff
@@ -525,14 +523,14 @@ pub fn from_yaml_str<T: Settings>(data: impl AsRef<str>) -> BootstrapResult<T> {
             },
         );
 
-        return match (path_res, saphyr_res) {
```
TheJokr, line 528: is there a lint that triggers here? i made the return explicit to ensure there is no fallthrough to the not-serde-saphyr block

orium: Yes: `clippy::needless_return`. But given the conditional compilation, it's safer to keep it. I'll `#[allow]` it here.

Cargo.toml
```diff
@@ -36,85 +36,88 @@ check-cfg = [
...
+tempfile = "3.27.0"
+tokio = "1.52.1"
+thread_local = "1.1.9"
+tikv-jemallocator = "0.6.1"
+tikv-jemalloc-ctl = "0.6.1"
+tower-service = "0.3.3"
+tracing-slog = "0.4.0"
+tracing-subscriber = "0.3.23"
+sentry-core = { version = "0.47.0", default-features = false }
+sentry = { version = "0.47.0", default-features = false, features = [
```
TheJokr: upgrading sentry is a breaking change, since we expose a sentry hook. this needs to stay at 0.36 until the next foundations major release.

E: added to tracking issue here https://github.com/cloudflare/foundations/issues/158#issuecomment-4327479343

orium: reverted back to 0.36.0

TheJokr (approved):

foundations/Cargo.toml
oga35767-eng: 👍

oga35767-eng (approved):

Cargo.toml
oga35767-eng: 👍

PR #203, opened 2026-04-27
Extract sentry hook into separate foundations-sentry crate

TheJokr:
Sentry's Rust client regularly publishes new major versions. They are on
version `0.47` now. Any `sentry` upgrade requires a `foundations` major
release as well, because we depend on sentry's public API. By extracting
the `sentry` feature from `foundations`, we can limit these major
releases to the new crate.

The new crate will use the latest sentry version from the start. We keep
the old code around for compatibility, but mark it as deprecated. It
will be removed in the next major release.

There are some alternatives:
- Allowing multiple major versions in our `Cargo.toml`. This is annoying
  because Cargo defaults to the latest version, causing duplicate
  dependencies and version mismatch errors.
- Having separate features per `sentry` major release. This would mean
  we need to maintain (or at least keep around) multiple versions of the
  same code, when really we just want people to upgrade to the latest
  sentry release.

Includes setup for releasing `foundations-sentry` separately from the main
crate and a release commit for `foundations-sentry 1.0.0`.


foundations-sentry/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+//! ```
+
+pub mod metrics;
+
+mod hook;
+mod settings;
+
+pub use self::hook::install_hook_with_settings;
+pub use self::settings::SentrySettings;
+pub use sentry_core::Level;
```
fisherdarling: Should we also `pub use sentry_core`? It would be convenient for users to pull in sentry through this crate. Since it's practically a wrapper around the sentry crate.

foundations-sentry/src/lib.rs
```diff
@@ -0,0 +1,43 @@
+#![allow(clippy::needless_doctest_main, reason = "for illustration")]
+//! Hook for rate-limiting sentry events and tracking them with [`foundations`] metrics.
```
fisherdarling: minor nit: let's prefix this with some kind of more official sounding language. Like "foundations sentry integration. Includes rate-limiting of sentry events and tracking them with [`foundations`] metrics." or smth (idk).

fisherdarling (approved):

foundations-sentry/src/lib.rs
TheJokr: IMO it's not a wrapper, more a plug-in like `sentry-anyhow`, `sentry-tower`, etc. You pass `ClientOptions` into the library to be modified. To initialize sentry you need `sentry::init` proper, rather than `sentry-core`. I'd prefer to leave this up to the user.

TBH, I don't think we should reexport `Level` either, it's only used for the metrics label.

fisherdarling: Makes sense to me. Good to know that's pretty normal in the `sentry-*` ecosystem.

PR #204, opened 2026-05-11
Allow pattern-based routes in TelemetryServer

TheJokr:
Routing is provided by `matchit`. By default (without the optional
`telemetry-server-pattern-routing` feature enabled) we escape all
parameter delimiters, which results in exact match routing as we had
before.

With the feature enabled, we pass the routes to `matchit` as-is,
allowing users to define routes with dynamic segments in them.

Includes a release commit for `foundations 5.7.0`.

### DIFF
--- a/foundations/src/telemetry/server/router.rs
+++ b/foundations/src/telemetry/server/router.rs
@@ -2,6 +2,7 @@
 use super::memory_profiling;
 #[cfg(feature = "memory-profiling")]
 use super::pprof_symbol;
+use crate::BootstrapResult;
 #[cfg(feature = "metrics")]
 use crate::telemetry::metrics;
 use crate::telemetry::reexports::http_body_util::{BodyExt, Empty, Full, combinators::BoxBody};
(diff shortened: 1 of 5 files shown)


foundations/src/telemetry/server/router.rs
```diff
@@ -31,8 +32,44 @@ pub type TelemetryRouteHandler = Box<
...
+        + Sync
+        + 'static,
+>;
 
 /// A telemetry server route descriptor.
+///
+/// There can be only one route handler per (Method, Path) pair. If there is a
+/// collision, the first route to be inserted wins. This includes built-in routes,
+/// which are inserted before any custom routes. New built-in routes may be added
+/// from time to time.
```
fisherdarling: Can we document the current built-in routes here?

I think it's
- `/health`
- `/metrics`
- `/pprof/heap`
- `/pprof/heap_stats`
- `/debug/traces`

Also, I wonder if we should add a blanket reservation like: "`/foundations/*` is reserved for future use" (perhaps `/internal/*`. Then we can be a bit more cavalier with adding new routes.

> or even better, `/cdn-cgi/foundations` :P

fisherdarling (approved):

PR #206, opened 2026-05-14
feat: add LogOutput::Custom variant for user-provided slog drains

ksurent:
Add a Custom variant to LogOutput that accepts an Arc<dyn CustomDrain>.  This allows users to provide their own slog drain (e.g., for Sentry integration) without needing wrapper macros or other workarounds.

The custom drain:
- Receives the same downstream wrapping as built-in drains (field dedup, field redaction, etc.)
- Is not serializable (must be set in code, not YAML)
- Ignores LogFormat (responsible for its own formatting)

The Custom variant uses Arc<dyn ...> instead of Box<dyn ...> because:

1. LogOutput must be Clone because LoggingSettings is cloned in telemetry::init(). Box<dyn ...> is not Clone.

2. telemetry::init() receives &LoggingSettings. A Box cannot be moved out of a shared reference, but an Arc can be cheaply cloned from one. Changing init() to take ownership would be a breaking API change.

Co–authored–by: Claude Opus 4.6

### DIFF
--- a/foundations/src/telemetry/log/testing.rs
+++ b/foundations/src/telemetry/log/testing.rs
@@ -93,6 +93,32 @@ pub(crate) fn create_test_log(
     (log, log_records)
 }
 
+/// Create a test log with a custom drain installed in addition to the test drain
+pub(crate) fn create_test_log_for_custom(
+    settings: &LoggingSettings,
+    custom_drain: crate::telemetry::settings::CustomDrain,
+) -> (LoggerWithKvNestingTracking, TestLogRecords) {
+    let log_records = Arc::new(RwLock::new(vec![]));
+
+    let drain = TestLogDrain {
+        records: Arc::clone(&log_records),
+        forward: Some(custom_drain),
+    };
+    let drain = wrap_root_drain(settings, drain);
+
+    let logger = build_log_with_drain(settings.verbosity, slog::o!(), Arc::clone(&drain));
+    let log = LoggerWithKvNestingTracking::new(logger);
+
+    let _ = LogHarness::override_for_testing(LogHarness {
+        root_log: Arc::new(ParkingRwLock::new(log.clone())),
+        root_drain: drain,
+        settings: settings.clone(),
+        log_scope_stack: Default::default(),
+    });
+
+    (log, log_records)
+}
+
 /// Create a test log with a tracing-rs compat drain installed in addition to the test drain
 #[cfg(feature = "tracing-rs-compat")]
 pub(crate) fn create_test_log_for_tracing_compat(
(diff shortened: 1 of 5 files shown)


foundations/src/telemetry/log/testing.rs
```diff
@@ -93,6 +93,32 @@ pub(crate) fn create_test_log(
     (log, log_records)
 }
 
+/// Create a test log with a custom drain installed in addition to the test drain
+pub(crate) fn create_test_log_for_custom(
+    settings: &LoggingSettings,
+    custom_drain: crate::telemetry::settings::CustomDrain,
```
TheJokr: please DRY: combine `create_test_log` and `create_test_log_for_custom` into a single function that takes `forward: Option<D>` where D: Drain.

foundations/src/telemetry/settings/logging.rs
```diff
@@ -3,6 +3,16 @@ use crate::utils::feature_use;
...
 
+#[cfg(feature = "logging")]
+use std::sync::Arc;
+
+#[cfg(feature = "logging")]
+use slog::{Never, SendSyncRefUnwindSafeDrain};
+
+/// A custom slog drain for use with [`LogOutput::Custom`].
+#[cfg(feature = "logging")]
+pub type CustomDrain = Arc<dyn SendSyncRefUnwindSafeDrain<Ok = (), Err = Never>>;
```
TheJokr: use `feature_use!` (as below) to avoid repeating `#[cfg(feature = "logging")]` three times

foundations/tests/custom_drain.rs
```diff
@@ -0,0 +1,42 @@
+use std::sync::{Arc, Mutex};
+
+use foundations::telemetry::log::error;
+use foundations::telemetry::settings::LoggingSettings;
+use foundations::telemetry::TestTelemetryContext;
+use foundations_macros::with_test_telemetry;
```
TheJokr: never import from foundations_macros directly, use `foundations::telemetry::with_test_telemetry`

ksurent: Fixed.

foundations/src/telemetry/settings/logging.rs
ksurent: Fixed.

foundations/src/telemetry/log/testing.rs
ksurent: Fixed. Simplified `create_test_log_for_tracing_compat` to use  `create_test_log` as well.

TheJokr (approved):

TheJokr:
Thank you!

PR #210, opened 2026-05-28
Add compile time check to make sure that types used as labels for metric families have the required traits

antoniovicente:
Use of types without the required traits will trigger an prometheus encoding failure so it's important to catch this kind of issue early.

### DIFF
--- a/foundations-macros/src/metrics/mod.rs
+++ b/foundations-macros/src/metrics/mod.rs
@@ -371,6 +371,10 @@ fn metric_fn(foundations: &Path, metrics_struct: &Ident, fn_: &ItemFn) -> proc_m
 
     let fn_args: Vec<_> = args.iter().map(|arg| arg.to_arg()).collect();
 
+    // Generate compile-time trait bound check for Family label types
+    let trait_bound_check =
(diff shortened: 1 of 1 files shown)


antoniovicente:
@TheJokr

TheJokr:
Thanks for the PR! I'm closing this since a separate trait check isn't necessary. Our macro generates a `registry.register(name, help, metric)` call which forces the compiler to do the necessary type checking already.

The problem you encountered at runtime is caused by the runtime behavior of the `serde::Serializer` that converts from the Family's label type to OpenMetrics format: https://github.com/nox/prometools/blob/v0.2.3/src/serde/mod.rs#L300
It returns an error if it encounters a type from serde's data model that doesn't directly map to a `key=value` syntax. With serde's design, AFAIK there is no compile-time check we could add for this.

I thought quite a bit about auto-generating a test to exercise encoding, but it would require a lot of boilerplate in foundations to be reliable. Also, as a user you wouldn't be able to see the source code for the test, which makes it hard to understand where the test even comes from. Instead, we changed our encoding to make errors non-fatal (#213). Errors will be logged for now, and once we implement #212 users will also be able to define a callback for other reporting (e.g., sentry).

PR #211, opened 2026-05-29
Add NoFlushPanicIntegration for sentry

TheJokr:
sentry's built-in panic handler integration explicitly flushes each event it generates. This makes sense as a default: a double-panic aborts the process, which would cause sentry event for the initial panic to be lost. However, it significantly increases the wall-clock duration of the panic handler if flushing means sending an HTTP request to a remote sentry server.

`NoFlushPanicIntegration` uses `sentry-panic`'s public API to generate the exact same panic events as sentry's built-in handler, but does not flush to speed up panic handler processing. This is preferrable for services where panics are caught and contained, for example in an HTTP server's request handler.


TheJokr:
cc @gregor-cf

fisherdarling (approved):

PR #213, opened 2026-06-02
Gracefully degrade on metrics encoding failures

TheJokr:
By default, `prometheus-client` aborts metrics collection/encoding on _any_ error. This means one error from the EncodeMetric implementation of a single metric in a Registry makes all the metrics in that Registry unavailable.

To avoid this, we introduce an EncodeMetric wrapper that swallows errors (after reporting them via logging/`eprintln!`.) Due to the way `prometheus-client` is designed, this could leave a partially-written metrics line in the output. We fix this by introducing `RewindableWriter`, a `Write` implementation that we can rewind to the last newline in case we do see an error.

`prometheus-client` does not give us access to the underlying writer in EncodeMetric impls, so we are forced to use a side channel via thread-local storage to activate the rewind behavior.

### DIFF
--- a/foundations/src/telemetry/metrics/internal.rs
+++ b/foundations/src/telemetry/metrics/internal.rs
@@ -1,7 +1,9 @@
-use super::{ExtraProducer, InfoMetric, info_metric};
+use super::rewind::{RewindState, RewindTo};
+use super::{ExtraProducer, InfoMetric, info_metric, report_nonfatal_collect_error};
 use crate::telemetry::settings::{MetricsSettings, ServiceNameFormat};
 use crate::{Result, ServiceInfo};
-use prometheus_client::encoding::text::{EncodeMetric, encode};
+use prometheus_client::encoding::text::{EncodeMetric, Encoder, SendSyncEncodeMetric, encode};
+use prometheus_client::metrics::MetricType;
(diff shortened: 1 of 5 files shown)


foundations/src/telemetry/metrics/internal.rs
```diff
@@ -176,19 +179,59 @@ where
...
-    encode(buffer, registry)?;
+std::thread_local! {
+    static ENCODER_REWIND_STATE: RewindState = const { RewindState::new() };
+}
+
+struct RewindErrorEncode<M>(M);
+
+impl<M: EncodeMetric> EncodeMetric for RewindErrorEncode<M> {
+    fn encode(&self, encoder: Encoder) -> std::io::Result<()> {
+        let mut res = self.0.encode(encoder);
```
fisherdarling, line 190: Should we `catch_unwind` here?

fisherdarling (approved):

foundations/src/telemetry/metrics/internal.rs
TheJokr: Discussed in dms: Catching panics here would require both `Encoder` and `M` to be UnwindSafe, which we can't guarantee. For example, prometool's `Family` type is `!UnwindSafe`.

PR #218, opened 2026-06-19
Drop Default supertrait from Settings

TheJokr:
This allows types without a sensible default value to implement `Settings`. Top-level setting types used with `Cli` still need to implement Default for the `--generate` option, but individual fields can now use types like `NonZeroUsize` or `SystemTime`. The `#[settings]` proc-macro still generates a serde-based Default impl. If a field does not have a Default impl, use `#[settings(impl_default = false)]` or specify a _field-specific_ Default with the `#[serde(default = ...)]` field attribute.

This is a breaking change: a function taking a `T: Settings` and calling `T::default()` was valid before, and is not anymore. However, this is the only breakage this change can cause.

Removing the Default supertrait also allows us to clean up the compatibility hack from foundations#150.


oga35767-eng (approved):

oga35767-eng:
👍

PR #223, opened 2026-06-24
User Tracing

mar-cf:
## Overview

This adds a **user-facing span pipeline** to `foundations`, parallel to the existing internal tracing pipeline. Application code can emit spans into a separate `USER_HARNESS` that exports **OTLP HTTP over a Unix domain socket** (gRPC unsupported) to an OTLP endpoint, with per-trace **routing metadata** (an application-defined `RoutingMetadata` trait) carried on the wire and W3C `traceparent` continuation in/out.

The design is deliberately a **mirror of internal tracing onto a second harness**: the user API lives in a dedicated `tracing::user_tracing` module whose entry points (`user_tracing::start_trace`, `user_tracing::span`, `user_tracing::add_span_tags!`, …) mirror the top-level internal-tracing functions (`start_trace`, `span`, `add_span_tags!`, …), plus a top-level `dual_span` that records into both pipelines at once. Two hard rules shape the surface:

- **The user and internal pipelines are independent** — separate harnesses, scope stacks, and exporters  
- **Public surface speaks W3C.**

There are two layers to keep separate:

- **init-time:** stand up `USER_HARNESS` \+ the exporter, pointed at the OTLP endpoint's socket.  
- **per-request activation:** `user_tracing::start_trace(...)` opens a root. Without a root, every `user_tracing::span` / `dual_span()` / `#[span_fn(user = true)]` / `user_tracing::add_span_tags!` is a **no-op**.

---

### Settings (init-time)

User tracing is configured via one optional block, gated by the `user-tracing` feature and fed to the existing `foundations::telemetry::init`:

```
TelemetrySettings.user_tracing: Option<UserTracingSettings>   // None = off
```

```rust
UserTracingSettings {
    enabled: bool,                       // default: false  (user tracing is opt-in)
    max_queue_size: Option<NonZeroUsize>,// default: 1_000_000; None = unbounded (mirrors internal)
    output: UserTracesOutput,            // only variant today: OtlpUds(..)
}

OtlpUdsOutputSettings {
    socket_path: String,        // ← path to the OTLP endpoint's UDS (required, no default)
                                //   e.g. "/path/to/otlp-receptor.sock"
    routing_header_name: String,// ← header the endpoint reads each span's encoded routing from
                                //   (required, no default) e.g. "cf-trace-config"
    num_tasks: usize,           // default: 1    (concurrent export workers)
    max_batch_size: usize,      // default: 512  (spans drained per export batch)
}
```

The two settings a consumer must choose are **`socket_path`** (the OTLP endpoint's UDS) and **`routing_header_name`** (the header the endpoint reads routing from). Unlike internal tracing, **user tracing is off by default** (`enabled: false`) and must be turned on explicitly; `max_queue_size` mirrors its internal-tracing equivalent; `num_tasks` and `max_batch_size` tune the exporter. All others have defaults and can be configured as needed.

There is deliberately **no sampling configuration** here. Unlike internal tracing, the user pipeline is not sampled inside foundations — the inbound `user_tracing` control header drives the activation (and therefore sampling) decision upstream.

### Instrumentation APIs

Toy example covering the whole surface:

```rust
use foundations::telemetry::tracing::{self, RoutingMetadata, TraceparentContext};

// `RoutingMetadata` is a trait — you define the concrete routing type. `group_key` batches
// spans that share it into one request; `encode` produces the routing header's value.
#[derive(Debug)]
struct MyRouting { zone_id: u64, account_id: u64 }

impl RoutingMetadata for MyRouting {
    fn group_key(&self) -> String { format!("{}|{}", self.zone_id, self.account_id) }
    fn encode(&self) -> String { format!("zone={};account={}", self.zone_id, self.account_id) }
}

// Per request: open the root. `routing` is required and fixed at construction (inherited by all
// descendants); `inbound` continues an upstream W3C trace, or None for a fresh one.
let _root = tracing::user_tracing::start_trace(
    "example_span_name",
    MyRouting { zone_id, account_id },
    inbound,                       // Option<TraceparentContext>
);

// Children — pick whichever fits:
let _child = tracing::user_tracing::span("lookup"); // explicit standalone user child
let _s     = tracing::dual_span("db");              // parallel internal + user child

#[tracing::span_fn("handle_request", user = true)]  // whole-fn dual span (sync or async)
async fn handle_request() { /* ... */ }

// Annotate the current user span (no-op if no user trace is active):
tracing::user_tracing::add_span_tags!("cache.status" => "HIT");

// Across .await / spawn / separate hooks (UserSpanScope is !Send): hold the context.
let ctx = tracing::user_tracing::start_trace("req", routing, None).into_context(); // -> TelemetryContext
ctx.apply(async { let _c = tracing::user_tracing::span("work"); }).await;

// Outbound propagation to the next hop:
let traceparent: Option<String> = tracing::user_tracing::w3c_traceparent();

// Inbound parsing (strict W3C):
let inbound = TraceparentContext::parse(header_bytes); // -> Option<TraceparentContext>
```

Key points for users:

- The **context carries the user span**: `into_context()` / `TelemetryContext::current()` / `#[span_fn]` all propagate it across `.await`, even through an *internal* span's context — no manual threading.  
- **`dual_span()` and `#[span_fn(user = true)]` open a user span in parallel with an internal span** — this matches the guideline to create an internal span for every user span, so a single call feeds both pipelines. `dual_span` names both spans from the same argument, so the user span keeps its name even when the internal trace is sampled out (the two pipelines sample independently).
- **Forking a trace leaves the user trace intact** — `TelemetryContext`'s trace fork only forks the *internal* trace; user spans created in the forked context stay in the current user trace.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ foundations = { version = "5.7.4", path = "./foundations", default-features = fa
 foundations-macros = { version = "=5.7.4", path = "./foundations-macros", default-features = false }
 bindgen = { version = "0.72.1", default-features = false }
 cc = "1.2.61"
-cf-rustracing = "1.3.0"
+cf-rustracing = "1.4.0"
 cf-rustracing-jaeger = "1.3.0"
 clap = "4.6.1"
 crossbeam-utils = { version = "0.8.21", default-features = false }
@@ -53,6 +53,7 @@ darling = "0.23.0"
 erased-serde = "0.4.10"
 futures-util = "0.3.32"
 governor = "0.10.4"
+hex = "0.4.3"
 http = "1.4.0"
 http-body-util = "0.1.3"
 hyper = { version = "1.9.0", default-features = false }
--- a/foundations/src/telemetry/settings/user_tracing.rs
+++ b/foundations/src/telemetry/settings/user_tracing.rs
@@ -1,0 +1,114 @@
+use std::num::NonZeroUsize;
+
+#[cfg(feature = "settings")]
+use crate::settings::settings;
+
+/// Distributed user tracing settings.
+#[cfg_attr(feature = "settings", settings(crate_path = "crate"))]
+#[cfg_attr(not(feature = "settings"), derive(Clone, Debug, serde::Deserialize))]
+pub struct UserTracingSettings {
+    /// Enables user tracing.
+    #[serde(default = "UserTracingSettings::default_enabled")]
+    pub enabled: bool,
+
+    /// Maximum number of spans to buffer for output. Any spans above
+    /// this limit will be dropped until the queue regains capacity.
+    ///
+    /// The default is to buffer up to 1 million spans in memory. This protects
+    /// services from out-of-memory errors when the output gets heavily backed up.
+    /// To disable the limit entirely, set this setting to `None`.
+    #[serde(default = "UserTracingSettings::default_max_queue_size")]
+    pub max_queue_size: Option<NonZeroUsize>,
+
+    /// The output for the collected user traces.
+    pub output: UserTracesOutput,
+}
+
+#[cfg(not(feature = "settings"))]
+impl Default for UserTracingSettings {
+    fn default() -> Self {
+        Self {
+            enabled: UserTracingSettings::default_enabled(),
+            max_queue_size: UserTracingSettings::default_max_queue_size(),
+            output: Default::default(),
+        }
+    }
+}
+
+impl UserTracingSettings {
+    fn default_enabled() -> bool {
+        false
+    }
+
+    const fn default_max_queue_size() -> Option<NonZeroUsize> {
+        Some(const { NonZeroUsize::new(1_000_000).expect("1_000_000 is not zero") })
+    }
+}
+
+/// The output for the collected user traces.
+#[cfg_attr(
+    feature = "settings",
+    settings(crate_path = "crate", impl_default = false)
+)]
+#[cfg_attr(not(feature = "settings"), derive(Clone, Debug, serde::Deserialize))]
+pub enum UserTracesOutput {
+    /// Send user tracing spans as OTLP over a Unix domain socket to an OTLP endpoint.
+    OtlpUds(OtlpUdsOutputSettings),
+}
+
+impl Default for UserTracesOutput {
+    fn default() -> Self {
+        Self::OtlpUds(Default::default())
+    }
+}
+
+/// [OTLP over UDS] output settings for user tracing.
+///
+/// Sends trace data as protobuf-encoded OTLP over HTTP to a Unix domain socket.
+#[cfg_attr(feature = "settings", settings(crate_path = "crate"))]
+#[cfg_attr(not(feature = "settings"), derive(Clone, Debug, serde::Deserialize))]
+pub struct OtlpUdsOutputSettings {
+    /// Path to the Unix domain socket for the OTLP endpoint.
+    pub socket_path: String,
+
+    /// Name of the header carrying each span's encoded routing.
+    pub routing_header_name: String,
+
+    /// Number of concurrent worker tasks for user trace export.
+    ///
+    /// # Default
+    ///
+    /// Default value is `1`.
+    #[serde(default = "OtlpUdsOutputSettings::default_num_tasks")]
+    pub num_tasks: usize,
+
+    /// Maximum number of spans to drain per batch.
+    ///
+    /// # Default
+    ///
+    /// Default value is `512`.
+    #[serde(default = "OtlpUdsOutputSettings::default_max_batch_size")]
+    pub max_batch_size: usize,
+}
+
+#[cfg(not(feature = "settings"))]
+impl Default for OtlpUdsOutputSettings {
+    fn default() -> Self {
+        Self {
+            socket_path: String::new(),
+            routing_header_name: String::new(),
+            num_tasks: OtlpUdsOutputSettings::default_num_tasks(),
+            max_batch_size: OtlpUdsOutputSettings::default_max_batch_size(),
+        }
+    }
+}
+
+impl OtlpUdsOutputSettings {
+    const fn default_num_tasks() -> usize {
+        1
+    }
+
+    const fn default_max_batch_size() -> usize {
+        512
+    }
+}
--- a/foundations/src/telemetry/telemetry_context.rs
+++ b/foundations/src/telemetry/telemetry_context.rs
@@ -21,6 +21,15 @@ feature_use!(cfg(feature = "tracing"), {
     });
 });
 
+feature_use!(cfg(feature = "user-tracing"), {
+    use super::tracing::UserSpanScope;
+    use super::tracing::internal::current_user_span;
+
+    feature_use!(cfg(feature = "testing"), {
+        use super::tracing::testing::{UserTestTracerScope, current_user_test_tracer};
+    });
+});
+
 #[cfg(feature = "testing")]
 use super::testing::TestTelemetryContext;
 
@@ -98,8 +107,14 @@ pub struct TelemetryContext {
     #[cfg(feature = "tracing")]
     pub(super) span: Option<SharedSpan>,
 
+    #[cfg(feature = "user-tracing")]
+    pub(super) user_span: Option<SharedSpan>,
+
     #[cfg(all(feature = "tracing", feature = "testing"))]
     pub(super) test_tracer: Option<Tracer>,
+
+    #[cfg(all(feature = "user-tracing", feature = "testing"))]
+    pub(super) user_test_tracer: Option<Tracer>,
 }
 
 impl TelemetryContext {
@@ -112,8 +127,14 @@ impl TelemetryContext {
             #[cfg(feature = "tracing")]
             span: current_span(),
 
+            #[cfg(feature = "user-tracing")]
+            user_span: current_user_span(),
+
             #[cfg(all(feature = "tracing", feature = "testing"))]
             test_tracer: current_test_tracer(),
+
+            #[cfg(all(feature = "user-tracing", feature = "testing"))]
+            user_test_tracer: current_user_test_tracer(),
         }
     }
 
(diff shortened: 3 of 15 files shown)


Cargo.toml
```diff
@@ -121,3 +122,7 @@ tower = "0.5.3"
 # Only used in deprecated `foundations` code
 # TODO: remove before next major release
 sentry-core = { version = "0.36.0", default-features = false }
+
+# TEMPORARY
+[patch.crates-io]
+cf-rustracing = { git = "https://github.com/mar-cf/rustracing.git", branch = "user-tracing" }
```
mar-cf: https://github.com/cloudflare/rustracing/pull/11

mar-cf:
@inikulin PTAL

foundations/src/telemetry/tracing/mod.rs
```diff
@@ -728,6 +764,68 @@ macro_rules! __set_span_finish_callback {
     }};
 }
 
+/// Adds tags to the current user span. No-op when no user trace is active.
```
TheJokr, line 875: please add references to the non-user macros that have more exhaustive docs. (you don't have to duplicate them.)

foundations/src/telemetry/tracing/mod.rs
```diff
@@ -158,6 +158,11 @@ pub use foundations_macros::span_fn;
 pub struct SpanScope {
     span: SharedSpan,
     _inner: Scope<SharedSpan>,
+
+    #[cfg(feature = "user-tracing")]
+    user_span: Option<SharedSpan>,
+    #[cfg(feature = "user-tracing")]
+    _user_inner: Option<Scope<SharedSpan>>,
```
TheJokr: can this be an `Option<UserSpanScope>`?

Also: As far as I understand it, this is only used to enable a fluent API like `span(...).with_user_tracing().into_context().apply(...)`. can we use a separate type for that instead so we dont pay the storage overhead on every SpanScope?

foundations/src/telemetry/telemetry_context.rs
TheJokr, line 362: please mention here that this only forks the internal (non-user) trace

foundations/src/telemetry/tracing/mod.rs
```diff
@@ -179,6 +181,27 @@ impl SpanScope {
...
 
+    /// Opens a parallel user span (child of the current user span, named after this span) when a
+    /// user trace is active; otherwise a no-op. The user span shares this scope's lifetime.
+    #[cfg(feature = "user-tracing")]
+    pub fn with_user_span(mut self) -> Self {
+        if current_user_span().is_some() {
+            let name = self
+                .span
+                .inner
+                .with_read(|s| s.operation_name().to_string());
```
TheJokr: I think given how user-tracing is structured, there could be a user trace even if the regular trace was dropped due to sampling. In that case `.operation_name()` returns an empty string. How should we handle this?

foundations/src/telemetry/settings/user_tracing.rs
```diff
@@ -0,0 +1,114 @@
...
+            enabled: UserTracingSettings::default_enabled(),
+            max_queue_size: UserTracingSettings::default_max_queue_size(),
+            output: Default::default(),
+        }
+    }
+}
+
+impl UserTracingSettings {
+    fn default_enabled() -> bool {
+        true
```
TheJokr: please dont enable tracing by default. users have to configure a path for the UDS anyway, so they should also explicitly set `enable: true`.

we plan to turn off regular tracing by default as well: https://github.com/cloudflare/foundations/issues/158#issuecomment-4004655787

foundations/src/telemetry/settings/user_tracing.rs
```diff
@@ -0,0 +1,114 @@
...
+            routing_header_name: String::new(),
+            num_tasks: OtlpUdsOutputSettings::default_num_tasks(),
+            max_batch_size: OtlpUdsOutputSettings::default_max_batch_size(),
+        }
+    }
+}
+
+impl OtlpUdsOutputSettings {
+    const fn default_num_tasks() -> usize {
+        2
```
TheJokr: the default for our other exporters is 1, can you add a comment explaining why you chose a higher number for user tracing?

foundations/src/telemetry/tracing/init.rs
```diff
@@ -167,3 +174,59 @@ pub(crate) fn init(
...
+#[cfg(feature = "user-tracing")]
+fn create_tracer_and_span_rx_for_user(
+    settings: &UserTracingSettings,
+) -> BootstrapResult<(Tracer, SharedSpanReceiver)> {
+    // The user pipeline samples everything it receives — the sampling decision
+    // is made upstream at activation time, outside foundations.
+    let sampler = AllSampler.boxed();
+
+    if let Some(cap) = settings.max_queue_size {
+        let (consumer, span_rx) = super::channel::channel(cap);
```
TheJokr: This channel uses a hardcoded gauge to record its current size. This needs to be changed to allow a gauge to be passed in. It's straightforward, and I don't this PR to become even larger, so I'll put up a quick PR tomorow.

foundations/src/telemetry/tracing/output_otlp_uds.rs
```diff
@@ -0,0 +1,545 @@
...
+        Ok(Self {
+            socket_path: settings.socket_path.clone(),
+            routing_header: settings.routing_header_name.clone(),
+        })
+    }
+
+    /// Processes a single drained batch of spans: groups them by routing,
+    /// converts each to OTLP, and POSTs one request per group. Errors are
+    /// reported and do not abort the batch.
+    async fn process_batch(&self, service_info: &ServiceInfo, spans: Vec<FinishedSpan>) {
```
TheJokr: nit: please use `spans: &mut Vec<...>` here and `spans.drain(..)` in the loop to avoid reallocating the vec in do_export every time

foundations/src/telemetry/tracing/mod.rs
```diff
@@ -513,6 +522,28 @@ pub fn user_span(name: impl Into<Cow<'static, str>>) -> UserSpanScope {
     UserSpanScope::new(create_user_span(name))
 }
 
+/// Introspection and outbound-propagation helpers for the user-tracing pipeline.
+#[cfg(feature = "user-tracing")]
+pub mod user_tracing {
```
TheJokr: Would you be ok with moving the rest of the public freestanding functions in here as well? i mean:
- `start_user_trace` -> `user_tracing::start_trace`
- `user_span` -> `user_tracing::span` 
- and maybe the macros as well

foundations/src/telemetry/tracing/traceparent.rs
```diff
@@ -0,0 +1,267 @@
...
+//! [`w3c_traceparent`](crate::telemetry::tracing::user_tracing::w3c_traceparent): it parses the
+//! W3C `traceparent` carried in the user-tracing control header so an inbound trace can be
+//! continued by [`start_user_trace`](crate::telemetry::tracing::start_user_trace).
+
+/// Parsed components of a valid W3C `traceparent` header value.
+///
+/// Constructed via [`TraceparentContext::parse`]. Round-trippable via
+/// [`TraceparentContext::to_traceparent_string`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TraceparentContext {
```
TheJokr, line 19: can you explain why you decided not to use https://docs.rs/cf-rustracing-jaeger/latest/cf_rustracing_jaeger/span/struct.SpanContextState.html? 

it has a FromStr implementation to parse the context from a string, and we have the code in `w3c_traceparent()` to convert back into a string. what is missing, and could we more easily add it in cf-rustracing-jaeger?

foundations/src/telemetry/tracing/init.rs
TheJokr: #230

foundations/src/telemetry/settings/user_tracing.rs
mar-cf: Done in 5127a35 - enabled now defaults to false

foundations/src/telemetry/settings/user_tracing.rs
mar-cf: Done in 6c74e99 - dropped the default to 1 to match the other exporters

foundations/src/telemetry/tracing/mod.rs
mar-cf: Done in d69896a - moved the user-only API under `user_tracing`

foundations/src/telemetry/tracing/traceparent.rs
mar-cf: Because that is Jaeger format, not W3C

The design covered some motivation for that choice

foundations/src/telemetry/tracing/output_otlp_uds.rs
mar-cf: Done in 96a9287

foundations/src/telemetry/tracing/mod.rs
mar-cf: Good catch. Confirmed with a test in dee068a. Fixed in f5d502a by introducing `dual_span` 

The other solutions were less appealing.

foundations/src/telemetry/telemetry_context.rs
mar-cf: Done in b6a6519

foundations/src/telemetry/tracing/mod.rs
mar-cf: Done in d63a187 - went with the separate-type route you suggested.

foundations/src/telemetry/tracing/mod.rs
mar-cf: Done in 0015623

mar-cf:
Will update the PR description shortly

Cargo.toml
TheJokr: merged & published as `cf-rustracing 1.4.0` now.

TheJokr (commented):
LGTM now. Please rebase/squash or resolve conflicts with main, bump the cf-rustracing version, and add `PipelineType::User` for the metrics, then we can merge this.

foundations/src/telemetry/tracing/traceparent.rs
TheJokr: oh i see, jaeger uses : as separators and w3c uses dashes. thanks for pointing that out, they are very similar 😅

mar-cf:
@TheJokr All done, ready to go

TheJokr (approved):

PR #229, opened 2026-07-10
feat(metrics-registry): add core registration API

ethanolchik:
Introduce the registration layer on top of the protobuf data model:

- `EncodeMetric`: trait for metrics that encode themselves into `MetricFamily` messages (best-effort; an empty `Vec` is valid).
- `register`/`iter`: append-only global registry behind a `parking_lot::RwLock<Vec<Arc<Entry>>>`. `iter()` clones an `Arc` snapshot so iteration never holds the lock. Service-name handling is deferred to encode time so metrics can be registered before the service name is known.
- `IntoMetrics`: sealed conversion trait accepting a single boxed metric or a `Vec` of them.
- `MetricsIter`/`MetricRegistration`: point-in-time snapshot iterator pairing each metric with its `RegistrationMetadata` via accessors, keeping the internal `Entry` storage type private.
- `RegistrationMetadata`: `#[non_exhaustive]` builder for the `optional` and `unprefixed` flags.

Add the `parking_lot` dependency and re-export the new public API from `lib.rs`.

### DIFF
--- a/foundations-metrics-registry/src/iter.rs
+++ b/foundations-metrics-registry/src/iter.rs
@@ -1,24 +1,22 @@
-use std::sync::Arc;
-
 use crate::EncodeMetric;
 use crate::RegistrationMetadata;
 use crate::registry::Entry;
 
 /// A registered metric together with its metadata.
 ///
-/// Yielded by [`MetricsIter`]
-pub struct MetricRegistration {
-    entry: Arc<Entry>,
+/// Yielded by [`MetricsIter`].
+pub struct RegisteredMetric {
+    entry: Entry,
 }
 
-impl MetricRegistration {
+impl RegisteredMetric {
     /// The metadata supplied to [`register`](crate::register).
     pub fn metadata(&self) -> &RegistrationMetadata {
         &self.entry.metadata
     }
     /// The registered metric.
     pub fn metric(&self) -> &dyn EncodeMetric {
-        self.entry.metric.as_ref()
+        self.entry.metric
     }
 }
 
@@ -27,11 +25,11 @@ impl MetricRegistration {
 /// Metrics registered after [`iter`](crate::iter()) was called are not observed,
 /// and the registry lock is not held while iterating.
 pub struct MetricsIter {
-    entries: std::vec::IntoIter<Arc<Entry>>,
+    entries: std::vec::IntoIter<Entry>,
 }
 
 impl MetricsIter {
-    pub(crate) fn new(entries: Vec<Arc<Entry>>) -> Self {
+    pub(crate) fn new(entries: Vec<Entry>) -> Self {
         Self {
             entries: entries.into_iter(),
         }
@@ -39,12 +37,12 @@ impl MetricsIter {
 }
 
 impl Iterator for MetricsIter {
-    type Item = MetricRegistration;
+    type Item = RegisteredMetric;
 
     fn next(&mut self) -> Option<Self::Item> {
         let entry = self.entries.next()?;
 
-        Some(MetricRegistration { entry })
+        Some(RegisteredMetric { entry })
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
--- a/foundations-metrics-registry/src/iter.rs
+++ b/foundations-metrics-registry/src/iter.rs
@@ -1,0 +1,55 @@
+use std::sync::Arc;
+
+use crate::EncodeMetric;
+use crate::RegistrationMetadata;
+use crate::registry::Entry;
+
+/// A registered metric together with its metadata.
+///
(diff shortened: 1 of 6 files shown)


foundations-metrics-registry/src/metadata.rs
```diff
@@ -0,0 +1,33 @@
...
+/// the setters, since downstream crates can't use a struct literal.
+#[non_exhaustive]
+#[derive(Clone, Default)]
+pub struct RegistrationMetadata {
+    /// Whether the metric is exported only when optional metrics are requested.
+    pub optional: bool,
+
+    /// Whether to suppress the service-name prefix for this metric
+    ///
+    /// The subsystem prefix stays; only the service-name prefix is skipped, and
```
TheJokr: no need to mention the subsystem prefix here IMO, this is something specific to how the macro works. if you registered a metric manually, we would never apply a subsystem prefix.

foundations-metrics-registry/src/metadata.rs
```diff
@@ -0,0 +1,33 @@
...
+/// `#[non_exhaustive]` so fields can be added later on without breaking
+/// [`register`](crate::register). Build it from [`default`](Self::default) plus
+/// the setters, since downstream crates can't use a struct literal.
+#[non_exhaustive]
+#[derive(Clone, Default)]
+pub struct RegistrationMetadata {
+    /// Whether the metric is exported only when optional metrics are requested.
+    pub optional: bool,
+
+    /// Whether to suppress the service-name prefix for this metric
```
TheJokr:
```suggestion
    /// Whether to suppress the service-name prefix for this metric.
```

foundations-metrics-registry/src/registry.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use crate::EncodeMetric;
+use crate::RegistrationMetadata;
+use crate::iter::MetricsIter;
+
+/// A registered metric paired with its registration metadata
+pub(crate) struct Entry {
+    pub(crate) metadata: RegistrationMetadata,
+    pub(crate) metric: Box<dyn EncodeMetric>,
+}
```
TheJokr, line 14: since we dont allow metrics to be unregistered, we can optimize this a bit by `Box::leak`ing the dyn objects:

```suggestion
#[derive(Clone)]
pub(crate) struct Entry {
    pub(crate) metadata: RegistrationMetadata,
    pub(crate) metric: &'static dyn EncodeMetric,
}
```

then we don't need the extra `Arc<...>` around Entry (saving 1 layer of indirection and making the vec clone a trivial bytewise copy).

foundations-metrics-registry/src/registry.rs
```diff
@@ -0,0 +1,108 @@
...
+pub fn iter() -> MetricsIter {
+    let entries: Vec<Arc<Entry>> = registry().read().clone();
+    MetricsIter::new(entries)
+}
+
+mod private {
+    pub trait Sealed {}
+}
+
+/// Converts a value into the metrics it contributes to the registry.
```
TheJokr, line 51:
```suggestion
/// Converts a value into the metrics it contributes to the registry.
///
/// This allows passing both `Box<dyn EncodeMetric>` and `Vec<Box<dyn EncodeMetric>>` to [`register`].
```

foundations-metrics-registry/src/iter.rs
```diff
@@ -0,0 +1,55 @@
+use std::sync::Arc;
+
+use crate::EncodeMetric;
+use crate::RegistrationMetadata;
+use crate::registry::Entry;
+
+/// A registered metric together with its metadata.
+///
+/// Yielded by [`MetricsIter`]
+pub struct MetricRegistration {
```
TheJokr: nit: i'd call this `RegisteredMetric`

(but please keep this wrapper when changing `Entry` so we dont expose the Clone impl)

ethanolchik: I agree with this change. In the spec I already defined a `RegisteredMetric` struct that stores the metric alongside its `name` and `help` attributes, but a better name for it is probably `NamedMetric`.

foundations-metrics-registry/src/registry.rs
ethanolchik: Ah yeah, then we can just hand out `&'static` refs. this is cool.

TheJokr: i'd actually keep the regular lifetime on the iterator accessor, in case we do want to change the storage back to a box or arc later

ethanolchik: ok makes sense

TheJokr (approved):

PR #231, opened 2026-07-14
Make stdio logging fail-soft on write errors

dineshadhi:
Foundations currently wraps terminal/stderr logging drains with `Fuse` before passing them to `slog_async`. If the underlying stdout/stderr stream returns an I/O error such as `BrokenPipe`, `Fuse` turns that error into a panic in the async logging worker.

This is unsafe for services whose stdout/stderr is captured by systemd and journald: if the journald stream is interrupted, normal logging can crash the service.

Handle terminal and stderr output errors with `ignore_res()` instead, so stdio logging failures do not panic the async logging worker. File logging keeps the existing fused behavior.

Add a regression test proving a failing async stdio drain keeps processing subsequent records instead of terminating after the first write error.

### DIFF
--- a/foundations/src/telemetry/log/init.rs
+++ b/foundations/src/telemetry/log/init.rs
@@ -13,8 +13,7 @@ use crate::telemetry::settings::{LogFormat, LogOutput, LogVerbosity, LoggingSett
 use crate::{BootstrapResult, ServiceInfo};
 use crossbeam_utils::CachePadded;
 use slog::{
-    Discard, Drain, FnValue, Fuse, Logger, OwnedKV, SendSyncRefUnwindSafeDrain,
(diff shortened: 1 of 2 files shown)


TheJokr:
The use of `fuse()` for our loggers is intentional - keeping a service running without log output is bad for observability. That being said, I understand there is a use case for this. I'm happy to merge a PR that makes this configurable:

- Add a field `ignore_io_errors: bool` to `LoggingSettings`
- Wrap the `AsyncDrain` construction in a new function that takes `drain: impl Drain<...>` and `settings: &LoggingSettings`
- Apply `fuse()` or `ignore_res()` depending on the setting there
- Use that function for the output types that are currently fused (terminal/stderr/file)

TheJokr (approved):

PR #232, opened 2026-07-14
feat(metrics): introduce foundations-metrics with scalar metrics

ethanolchik:
# Overview

This pull request introduces the new `foundations-metrics` crate, which provides the initial scalar metric types and converts their values into the Prometheus protobuf data model.

It builds on `foundations-metrics-registry`, which continues to own the stable process-global registry and protobuf types. This separation allows metric implementations to evolve without destabilising the registry shared across Foundations versions.

## New Crate: foundations-metrics

* Added the crate's public API, module structure, and documentation.
* Re-exported `register`, `EncodeMetric`, `IntoMetrics`, `MetricFamily`, and `RegistrationMetadata`, allowing callers to use the higher-level crate without importing registry APIs separately.

## Scalar Metrics

* Implemented `Counter` with `u64` and `f64` support.
* Implemented `Gauge` with `i64` (since `prometheus_client` v0.19.0), `u64`, and `f64` support.
* Implemented `RangeGauge`, which exports the current, minimum, and maximum values observed since the previous collection.
* Implemented `GaugeGuard`, which increments a gauge when created and decrements it when dropped.
* Scalar metrics own their atomic storage rather than wrapping `prometheus-client`. The spec allowed either approach; owned storage keeps the scalar API independent of upstream implementation details and defers the dependency until it is needed for native histograms, though it is still included in `foundations-metrics/Cargo.toml`.
* Cloned metrics share their underlying atomic storage through `Arc`, preserving inexpensive handle semantics.
* Built-in storage intentionally focuses on the 64-bit numeric types currently used by Foundations. The public `CounterAtomic` and `GaugeAtomic` traits allow additional storage implementations to be added without changing the metric API.

## Registration and Encoding

* Introduced `NamedMetric`, which associates metric storage with its exported name and help text. Metric handles intentionally contain only state because the metrics macro already owns registration metadata.
* Added the internal `EncodeMetricValue` trait. Storage types encode protobuf families with relative names, while `NamedMetric` applies the registered base name and help text.
* Scalar metrics emit an empty relative suffix. `RangeGauge` emits `""`, `_min`, and `_max`, allowing it to compose with `NamedMetric` and future labelled families without special handling.
* Counter families are encoded without automatically appending `_total`, preserving existing Foundations output behaviour.

## Workspace Updates

* Registered `foundations-metrics` as a workspace member and workspace dependency.
* Updated `foundations-metrics-registry` to export the protobuf `Counter` and `Gauge` message types.

### DIFF
--- a/foundations-metrics/Cargo.toml
+++ b/foundations-metrics/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "foundations-metrics"
+version.workspace = true
+repository.workspace = true
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+
+[dependencies]
+foundations-metrics-registry = { workspace = true }
+prometheus-client = "0.25.0"
+
+[lints]
+workspace = true
--- a/foundations-metrics/Cargo.toml
+++ b/foundations-metrics/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "foundations-metrics"
-version.workspace = true
+version = "0.1.0"
 repository.workspace = true
 edition.workspace = true
 authors.workspace = true
--- a/foundations-metrics/src/metrics/gauge.rs
+++ b/foundations-metrics/src/metrics/gauge.rs
@@ -1,10 +1,8 @@
-//! Gauge metrics and their lifecycle helpers.
-
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::marker::PhantomData;
+use std::sync::atomic::{AtomicI64, AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 
 use foundations_metrics_registry::proto::{self, MetricFamily, MetricType};
-use prometheus_client::metrics::gauge::{Atomic as GaugeAtomic, Gauge as PrometheusGauge};
 
 use crate::value::EncodeMetricValue;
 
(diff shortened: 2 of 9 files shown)


foundations-metrics/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "foundations-metrics"
+version.workspace = true
```
TheJokr:
```suggestion
version = "0.1.0"
```
versioning should be independent from the workspace version (which is for foundations & foundations-macros). just start with 0.1.0 for now.

foundations-metrics/src/registered.rs
```diff
@@ -0,0 +1,108 @@
...
+impl<M> EncodeMetric for NamedMetric<M>
+where
+    M: EncodeMetricValue,
+{
+    fn encode(&self) -> Vec<MetricFamily> {
+        let mut families = self.metric.encode_metric_value();
+
+        for family in &mut families {
+            let suffix = family.name.take().unwrap_or_default();
+            family.name = Some(format!("{}{}", self.name, suffix));
```
TheJokr:
```suggestion
            let name = family.name.get_or_insert_default();
            name.insert_str(0, self.name);
```
`insert_str` has 2 advantages:
- It potentially allows the allocation from `family.name` to be reused (if the String over-allocated)
- It immediately allocates the exact size that is needed, whereas `format!` desugars to 2 separate `<String as Display>::fmt` calls that may need separate allocations

foundations-metrics/src/metrics/mod.rs
```diff
@@ -0,0 +1,43 @@
+//! The metrics module that contains scalar metrics (like Counter, Gauge, ...) and non-scalars,
+//! like Histogram, NativeHistogram, and also Family.
```
TheJokr, line 2: a lot of this code is adapted/forked from prometheus-client right? i think its best if we put a disclaimer/attribution in the docs here. something like:

````md
# License
Substantial parts of the code in this module were adapted from [`prometheus-client`]
subject to their licensing terms:
```
<copy of https://github.com/prometheus/client_rust/blob/master/LICENSE-MIT>
```
````

File: foundations-metrics/src/metrics/gauge.rs [hunk 4]
```diff
@@ -0,0 +1,593 @@
...
+
+use super::IntoF64;
+
+/// A metric whose value may increase, decrease, or be set directly.
+///
+/// It is a cheap handle over shared atomic storage: [`Clone`] hands out another
+/// reference to the *same* series, so the gauge can be updated from many places
+/// and read back as a single current value.
+///
+/// The `u64`/`AtomicU64` defaults preserve the existing Foundations API.
```
TheJokr: nit: don't mention this in the public docs, if you want you can keep this as a regular non-doc comment

foundations-metrics/src/metrics/mod.rs
ethanolchik: Yeah great i'll do that. Also, in the next set of changes, some things are adapted from prometools too, but not sure if I should mention that

foundations-metrics/src/registered.rs
ethanolchik: Never knew this, very interesting!

foundations-metrics/src/metrics/mod.rs
TheJokr: (discussed in DMs, Ethan will add a note for prometools together with that code)

TheJokr (approved):

PR #234, opened 2026-07-16
feat(metrics): add support for metric families and label serialisation

ethanolchik:
## Overview

This PR builds on the already-merged scalar metrics #232  (`Counter`, `Gauge`) to add **labelled metrics**. It introduces a serde-based system for serialising metric label sets into the Prometheus protobuf data model, and adds a `Family` abstraction for grouping metrics by their label values.

## Label serialisation and error handling

* Introduced a new `labels` module with:
  * `to_label_pairs`: serialises any `serde::Serialize` label set into protobuf `LabelPair` values.
  * A serde `Serializer` implementation that flattens a label struct into name/value pairs, **rejecting** unsupported compound types (sequences, maps, nested structs, byte arrays) with clear errors.
  * `LabelError`: a dedicated error type implementing `serde::ser::Error`.
* Preserves the scalar formatting behaviour of the existing metrics backend: `ryu`-based float formatting, serde renames, enum variants, optional values, and `Display` adapters.
* Validates label names against `[a-zA-Z_:][a-zA-Z0-9_:]*`.
* Stores label values as raw strings - OpenMetrics escaping is deliberately deferred to the text encoder, keeping the protobuf model format-independent.
* Added `serde` (with `derive`), `ryu`, and `parking_lot` dependencies to support label serialisation and efficient concurrent metric storage.

## Metrics API enhancements

* Added `Family<S, M, C>`: shared, labelled metric storage with:
  * `get_or_create`, `remove`, and `clear` lifecycle operations; clones share the same underlying metrics.
  * A `MetricConstructor` trait (with a blanket `Fn() -> M` impl) for metrics that need configuration at creation time.
* **Fail-soft collection:** when encoding, invalid label sets and metric groups with inconsistent metadata are skipped (with a diagnostic) rather than failing the entire scrape; multi-series metrics (e.g. `RangeGauge`) are grouped correctly by suffix.
* Re-exported `Family`, `MetricConstructor`, `LabelError`, `to_label_pairs`, and `set_collect_error_hook` from the crate root for downstream use.

## Decoupling logging from metrics

* Collection diagnostics are emitted through an installable hook (`set_collect_error_hook`) rather than a hard-coded logger, defaulting to stderr when none is set.
* **Follow-up (for when `foundations` adopts this crate):** install a collect_error hook that forwards these diagnostics to the telemetry logging pipeline (at warning level), matching the legacy backend's behaviour.

## Protobuf model adjustments

* Re-exported `LabelPair` from the registry `proto` module so downstream code can construct/consume label pairs.

## Attribution

* The label serialiser and `Family` are adapted from [`prometools`](https://github.com/nox/prometools) (`MIT OR Apache-2.0`); per-definition `// Adapted from ...` notes were added, per the discussion with @TheJokr.

### DIFF
--- a/foundations-metrics/src/metrics/family.rs
+++ b/foundations-metrics/src/metrics/family.rs
@@ -162,6 +162,42 @@ where
         }))
     }
 
+    /// Returns an owned handle to the metric for `label_set`, creating it when
+    /// absent.
+    ///
+    /// The returned handle is a cheap clone that shares storage with the metric
+    /// in the family, so updates through it are reflected in collection. Unlike
+    /// [`get_or_create`](Self::get_or_create), it does not keep the family
+    /// locked, so it is safe to hold and reuse (and avoids that method's
+    /// deadlock hazard).
+    ///
+    /// For a stable label set, prefer resolving the handle once and reusing it
+    /// rather than calling the family on every update.
+    ///
+    /// ```
+    /// use foundations_metrics::{Counter, Family};
+    /// use serde::Serialize;
+    ///
+    /// #[derive(Clone, Eq, Hash, PartialEq, Serialize)]
+    /// struct Labels {
+    ///     method: &'static str,
+    /// }
+    ///
+    /// let requests = Family::<Labels, Counter>::default();
+    ///
+    /// // Resolve the handle once, then reuse it on the hot path.
+    /// let gets = requests.get_or_create_owned(&Labels { method: "GET" });
+    /// gets.inc();
+    /// gets.inc();
+    /// assert_eq!(gets.get(), 2);
+    /// ```
+    pub fn get_or_create_owned(&self, label_set: &S) -> M
+    where
+        M: Clone,
+    {
+        M::clone(&self.get_or_create(label_set))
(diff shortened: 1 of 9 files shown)


foundations-metrics/src/metrics/family.rs
```diff
@@ -0,0 +1,466 @@
...
+where
+    S: Clone + Eq + Hash,
+    C: MetricConstructor<M>,
+{
+    /// Returns the metric for `label_set`, creating it when absent.
+    ///
+    /// The returned guard keeps the family read-locked. Holding it while
+    /// accessing another label set in the same family can deadlock if that
+    /// second label set needs to be created.
+    pub fn get_or_create(&self, label_set: &S) -> MappedRwLockReadGuard<'_, M> {
```
fisherdarling: Can we create our own `ReadMetric` wrapper (not sure on the name 😅). I'd like to avoid having parking_lot in the public api. Since then that let's us experiment with internal locking mechanisms / sharding / w/e.

fisherdarling (approved):

PR #235, opened 2026-07-17
Remove seccomp probe results

patrickmscott:
Compile the seccomp header probe in a subdirectory and remove. The output is not used by the final crate. In build systems like bazel, the generated archive names are non-deterministic and produce different digests.


fisherdarling (approved):

patrickmscott:
@fisherdarling thanks for the quick review. Anything else I need to do to get this merged?

fisherdarling:
@patrickmscott nope! I just need to run the workflow and click the button, on it!

PR #236, opened 2026-07-20
feat(metrics): add support for classic, time, and native histograms.

ethanolchik:
## Overview

Adds classic, time, and native histogram metric types to `foundations-metrics`, including protobuf encoding, constructors, family support, `TimeHistogram` snapshots, and timer behaviour.

### Classic histograms

Adds an owned `Histogram` backed by shared, mutex-protected storage.

- `Histogram::new` accepts inclusive bucket bounds and appends the legacy terminal `f64::MAX` bucket.
- Cloned handles share observations.
- `observe` updates the sample sum, sample count, and the first matching bucket.
- Sorted buckets use a binary search while unsorted buckets retain the legacy linear-search behaviour.
- NaN observations preserve legacy behaviour by updating sum/count without incrementing a bucket.
- Protobuf output contains a histogram row with cumulative bucket counts, sample count, and sample sum.

### Time histograms

Adds the lock-free `TimeHistogram` and `HistogramTimer` APIs previously provided by `prometools`.

- Observations are stored atomically in nanoseconds and exported in seconds.
- `start_timer` records elapsed time when the timer is stopped or dropped.
- Timers support `pause`, `resume`, `stop_and_record`, and `stop_and_discard`.
- A timer records at most once.
- `TimeHistogram::snapshot` exposes the current sum, count, and non-cumulative buckets through `HistogramSnapshot`.

### Histogram construction

Adds `HistogramBuilder` as a `MetricConstructor` for both `Histogram` and `TimeHistogram`.

This allows configured histograms to be created by `Family` and by generated metric constructors while preserving the existing `#[ctor = HistogramBuilder { ... }]` shape.

### Native histograms

Adds `NativeHistogram`, backed by `prometheus-client` 0.25's native histogram implementation.

- `NativeHistogram::observe` records values into sparse exponential buckets.
- Cloned handles share the same storage.
- `NativeHistogramBuilder` configures bucket growth factor, zero-bucket threshold, and the best-effort maximum number of sparse buckets.
- The builder implements `MetricConstructor`, allowing native histograms to be used in labelled `Family` metrics.
- Native histogram encoding delegates to `prometheus-client` because its sparse bucket state is private.
- The resulting protobuf families are converted into the canonical `foundations-metrics-registry` model, including sample count/sum, schema, zero bucket, positive and negative spans, and bucket deltas.
- Encoding errors are reported through the existing non-fatal collection diagnostic hook and skip only the affected histogram.

### Public API

Exports the following from `foundations-metrics`:

- `Histogram`
- `HistogramBuilder`
- `HistogramSnapshot`
- `HistogramTimer`
- `TimeHistogram`
- `NativeHistogram`
- `NativeHistogramBuilder`

The registry crate also re-exports the protobuf `Bucket` type needed by histogram encoders.

### DIFF
--- a/foundations-metrics/src/metrics/histogram.rs
+++ b/foundations-metrics/src/metrics/histogram.rs
@@ -247,17 +247,21 @@ impl Clone for TimeHistogram {
 impl TimeHistogram {
     /// Creates a time histogram with inclusive bucket bounds in seconds.
     ///
-    /// A terminal `f64::MAX` bucket is appended automatically.
+    /// Bounds may be given in any order; they are sorted ascending. A terminal
+    /// `f64::MAX` bucket is appended automatically.
     pub fn new(buckets: impl IntoIterator<Item = f64>) -> Self {
+        let mut buckets: Vec<_> = buckets
+            .into_iter()
+            .chain(once(f64::MAX))
+            .map(|upper_bound| (upper_bound, AtomicU64::new(0)))
+            .collect();
+        buckets.sort_by(|(a, _), (b, _)| a.total_cmp(b));
+
         Self {
             state: Arc::new(TimeHistogramState {
                 sum: Default::default(),
                 count: Default::default(),
-                buckets: buckets
-                    .into_iter()
-                    .chain(once(f64::MAX))
-                    .map(|upper_bound| (upper_bound, AtomicU64::new(0)))
-                    .collect(),
+                buckets,
             }),
         }
     }
@@ -477,6 +481,17 @@ mod tests {
                 .map(|bucket| bucket.cumulative_count)
                 .collect::<Vec<_>>(),
             vec![Some(1), Some(2), Some(3)]
+        );
+    }
+
+    #[test]
+    fn time_histogram_sorts_unsorted_buckets() {
+        let histogram = TimeHistogram::new([2.0, 1.0]);
+        histogram.observe(500_000_000);
+
+        assert_eq!(
+            histogram.snapshot().buckets(),
+            &[(1.0, 1), (2.0, 0), (f64::MAX, 0)]
         );
     }
 
--- a/foundations-metrics/src/metrics/histogram.rs
+++ b/foundations-metrics/src/metrics/histogram.rs
@@ -290,7 +290,7 @@ impl TimeHistogram {
             .buckets
             .iter()
             .enumerate()
-            .find(|(_i, (upper_bound, _value))| upper_bound >= &(v as f64 * 1E-9));
+            .find(|(_i, (upper_bound, _value))| upper_bound >= &(seconds(v)));
 
         match first_bucket {
             Some((i, (_upper_bound, value))) => {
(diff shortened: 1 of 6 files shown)


foundations-metrics/src/metrics/histogram.rs
```diff
@@ -0,0 +1,251 @@
...
+impl Histogram {
+    /// Creates a histogram with the provided inclusive upper bounds.
+    ///
+    /// Bounds should be ordered from smallest to largest. A terminal
+    /// `f64::MAX` bucket is appended automatically.
+    pub fn new(bounds: impl IntoIterator<Item = f64>) -> Self {
+        let mut buckets: Vec<_> = bounds.into_iter().map(|bound| (bound, 0)).collect();
+        buckets.push((f64::MAX, 0));
+
+        let sorted = buckets.windows(2).all(|window| window[0].0 <= window[1].0);
```
fisherdarling: Can we pre-sort the given buckets, rather than storing a boolean?

foundations-metrics/src/metrics/histogram.rs
```diff
@@ -0,0 +1,251 @@
...
+#[derive(Clone, Debug)]
+pub struct Histogram {
+    state: Arc<Mutex<State>>,
+}
+
+#[derive(Debug)]
+struct State {
+    sum: f64,
+    count: u64,
+    buckets: Vec<(f64, u64)>,
```
fisherdarling, line 37: If we make these atomics, we can drop the `Mutex` around `State`, right?

fisherdarling: Ah, but then we lose consistency across `sum` and `count`...
Hm :/

edit: which is maybe ok? looks like that's the case for the TimeHistogram

foundations-metrics/src/metrics/histogram.rs
```diff
@@ -146,6 +152,193 @@ impl MetricConstructor<Histogram> for HistogramBuilder {
...
+                value.fetch_add(1, Ordering::Relaxed);
+                Some(i)
+            }
+            None => None,
+        }
+    }
+
+    /// Returns a snapshot whose sum and bucket bounds are expressed in seconds.
+    pub fn snapshot(&self) -> HistogramSnapshot {
+        let sum = (self.state.sum.load(Ordering::Relaxed) as f64) * 1E-9;
```
fisherdarling: small nit: can we make `1E-9` a named constant?

foundations-metrics/src/metrics/histogram.rs
ethanolchik: Right. I think `TimeHistogram` deliberately does this for its lock-free hot path, but it therefore loses consistency as you mentioned. I think keeping the Mutex is the safer choice here. We can profile later to see if the lock shows up and decide if this needs to change if that works for you?

foundations-metrics/src/metrics/histogram.rs
ethanolchik: yeah, you're right we can actually pre-sort. I initially thought this would break some legacy compatibility since users can create histograms with unsorted buckets with `Histogram::new` , but any data those produce is pretty much garbage anyways.

foundations-metrics/src/metrics/histogram.rs
ethanolchik: created the seconds() function, forked from prometools

foundations-metrics/src/metrics/histogram.rs
fisherdarling: Yea that's fine, we can improve perf as we go! Relaxing the consistency I guess is technically a minor breaking change.

fisherdarling (approved):

ethanolchik:
The latest changes make `Histogram::snapshot` public, matching the spec, and also fixing panic locations in native histograms.

PR #238, opened 2026-07-24
feat(metrics): add metric validation + UTF-8 support

ethanolchik:
## What Changed
Introduces shared validation used during collection and by both encoders.

Names are valid when they are non-empty UTF-8 strings without NUL (`\0`) bytes. Names containing spaces, punctuation, newlines, or non-ASCII characters remain valid and are escaped appropriately for text output.

Validation failures are isolated as narrowly as possible:
- An invalid family name drops that family.
- An invalid, duplicate, or reserved label drops only its metric row.
- An invalid or duplicate exemplar label drops only the exemplar.
    - This branch validates exemplars already present in `MetricFamily`; exemplar-producing metric types are introduced separately.
- Other families and rows continue to be collected and encoded.
- Every dropped value produces a non-fatal diagnostic with collection or encoding context.

Histogram and gauge-histogram rows reserve `le`, while summary rows reserve quantile, according to the canonical metric model.
## Service Labels
Service-label insertion now handles existing labels safely:
- A row already containing the configured label and value is unchanged.
- A row containing the same label with a different value is dropped.
- A missing service label is prepended as before.
- An invalid configured service-label name prevents collection instead of producing malformed output.
## OpenMetrics Text
The text encoder now:
- Quotes label names only whne they require UTF-8 syntax; legacy-compatible label names remain unquoted.
- Uses quoted OpenMetrics syntax for metric names outside the legacy grammar.
- Escapes metric and label names safely.
- Appends histogram suffixes inside quoted metric names.
- Exports `OPENMETRICS_CONTENT_TYPE` with `escaping=allow-utf-8`.
Legacy-compatible metric names remain unquoted.
## Protobuf
Protobuf output preserves valid UTF-8 names exactly. The protobuf encoder applies the same defensive validation when called directly, so bypassing collection does not bypass validation.

### DIFF
--- a/foundations-metrics/src/collect.rs
+++ b/foundations-metrics/src/collect.rs
@@ -1,6 +1,10 @@
 use foundations_metrics_registry::{iter, proto::LabelPair};
 
 use crate::MetricFamily;
+use crate::diagnostics::report_collect_error;
+use crate::validation::{
+    LABEL_NAME_GRAMMAR, ValidationContext, is_valid_label_name, sanitize_metric_family,
+};
 
 /// Options that control which registered metrics are collected and how the
 /// service name is represented.
@@ -28,6 +32,16 @@ pub enum ServiceNameFormat<'a> {
 
 /// Collects the currently registered metrics into the canonical protobuf model.
 pub fn collect(options: CollectionOptions) -> Vec<MetricFamily> {
+    if options.service_name.is_some()
+        && let ServiceNameFormat::LabelWithName(label_name) = options.service_name_format
+        && !is_valid_label_name(label_name)
+    {
+        report_collect_error(format_args!(
+            "non-fatal error while collecting metrics: invalid configured service label name {label_name:?}; expected {LABEL_NAME_GRAMMAR}; skipped all metric families"
+        ));
+        return Vec::new();
+    }
+
     let mut collected = Vec::new();
 
     for registered in iter() {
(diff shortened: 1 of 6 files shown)


foundations-metrics/src/collect.rs
```diff
@@ -56,15 +70,35 @@ pub fn collect(options: CollectionOptions) -> Vec<MetricFamily> {
...
+                            }
+                            true
+                        });
                     }
                 }
                 ServiceNameFormat::MetricPrefix => {}
             }
         }
 
+        families.retain_mut(|family| sanitize_metric_family(family, ValidationContext::Collection));
```
fisherdarling: I think we're doing some extra allocation/memcpy work here. Rather than `retain_mut`, we can move to a `filter_map` and extend the collected from the iter.

foundations-metrics/src/collect.rs
```diff
@@ -28,6 +32,16 @@ pub enum ServiceNameFormat<'a> {
 
 /// Collects the currently registered metrics into the canonical protobuf model.
 pub fn collect(options: CollectionOptions) -> Vec<MetricFamily> {
```
fisherdarling, line 34: nit: can we refactor this function a little bit? There's a fair bit of indentation. I think it's worthy of splitting some things out. Perhaps add a helper for modifying the `ServiceNameFormat::LabelWithName` branch. Maybe even move the service name appending code into a helper.

fisherdarling (approved):

PR #239, opened 2026-07-24
feat(metrics): Implement exemplar metrics

ethanolchik:
## Overview
Adds first-party exemplar-aware metric types for counters, classic histograms, and native histograms while leaving existing metric types unchanged.

## Public API
- `CounterWithExemplar`
- `HistogramWithExemplars`
- `NativeHistogramWithExemplars`
- Read-only `Exemplar` accessors for labels and sampled values are exposed through `CounterWithExemplar::get`.

These types integrate with `Family`, registration, collection, and the existing protobuf and OpenMetrics encoders.
## Recording Semantics
- Counters retain the exemplar from the latest labeled increment. An unlabelled increment clears it.
- Classic histograms retain the latest labeled observation for each bucket. Unlabelled observations do not clear existing exemplars.
- Native histograms retain the latest labeled observation and capture its timestamp.
- Clones share metric and exemplar storage.
- Family series labels remain separate from exemplar labels.
Recording does not require exemplar labels to implement `Serialize`; serialisation occurs when the metric is encoded.
## Encoding
Counter and classic-histogram exemplars are emitted in both protobuf and OpenMetrics text.
Native-histogram exemplars:
- Include the timestamp required by the Prometheus protobuf model.
- Are preserved in protobuf output.
- Are not emitted in text because native histograms require protobuf exposition.

Empty counter and classic-histogram exemplar label sets remain valid and are emitted in text as {}.

## Labels And Validation
Exemplar labels use the existing label serialiser and support:
- Serialisable structs
- Unit label sets
- Legacy sequences of name-value pairs

Invalid, duplicate, or unserialisable exemplar labels drop only the exemplar. The metric sample remains available, and a non-fatal diagnostic is emitted.

Native-histogram exemplars without the required timestamp are removed with a non-fatal diagnostic; the native histogram itself is still emitted.

### DIFF
--- a/foundations-metrics/src/metrics/counter.rs
+++ b/foundations-metrics/src/metrics/counter.rs
@@ -135,7 +135,7 @@ impl<N, A: Default> Default for Counter<N, A> {
 ///
 /// The name/help are left empty here; they are populated at registration and
 /// encode time.
-fn encode_counter(value: f64) -> Vec<MetricFamily> {
+pub(super) fn encode_counter(value: f64) -> Vec<MetricFamily> {
     vec![MetricFamily {
         name: Some(String::new()),
         help: None,
--- a/foundations-metrics/src/metrics/counter.rs
+++ b/foundations-metrics/src/metrics/counter.rs
@@ -3,8 +3,10 @@ use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
 
 use foundations_metrics_registry::proto::{self, MetricType};
+use serde::Serialize;
 
 use super::IntoF64;
+use super::exemplar::{Exemplar, ExemplarStorage, StoredExemplar, WithExemplar, finish_exemplar};
 
 use crate::{MetricFamily, value::EncodeMetricValue};
 
@@ -158,6 +160,81 @@ where
 {
     fn encode_metric_value(&self) -> Vec<MetricFamily> {
         encode_counter(self.get().into_f64())
+    }
+}
+
+macro_rules! impl_counter_with_exemplar {
+    ($value:ty, $variant:ident) => {
+        impl<S, A> WithExemplar<Counter<$value, A>, S>
+        where
+            A: CounterAtomic<$value>,
+        {
+            /// Returns the current exemplar.
+            pub fn exemplar(&self) -> Option<Exemplar<S, $value>> {
+                match &*self.exemplars.lock() {
+                    ExemplarStorage::Empty | ExemplarStorage::Single(None) => None,
+                    ExemplarStorage::Single(Some(StoredExemplar::$variant(exemplar))) => {
+                        Some(exemplar.clone())
+                    }
+                    ExemplarStorage::Single(Some(_)) => {
+                        unreachable!("counter uses matching exemplar value storage")
+                    }
+                    ExemplarStorage::PerBucket(_) => {
+                        unreachable!("counter uses single exemplar storage")
+                    }
+                }
+            }
+
+            /// Increments the counter by `value`, records the exemplar, and
+            /// returns the previous total.
+            pub fn with_exemplar(&self, label_set: S, value: $value) -> $value {
+                let exemplar = StoredExemplar::$variant(Exemplar::new(label_set, value, None));
+                let mut exemplars = self.exemplars.lock();
+                let previous = self.inner.inc_by(value);
+
+                match &mut *exemplars {
+                    ExemplarStorage::Empty => {
+                        *exemplars = ExemplarStorage::Single(Some(exemplar));
+                    }
+                    ExemplarStorage::Single(stored) => *stored = Some(exemplar),
+                    ExemplarStorage::PerBucket(_) => {
+                        unreachable!("counter uses single exemplar storage")
+                    }
+                }
+
+                previous
+            }
+        }
+    };
+}
+
+impl_counter_with_exemplar!(i64, I64);
+impl_counter_with_exemplar!(u64, U64);
+impl_counter_with_exemplar!(f64, F64);
+
+impl<S, N, A> EncodeMetricValue for WithExemplar<Counter<N, A>, S>
+where
+    S: Serialize + Send + Sync + 'static,
+    N: IntoF64,
+    A: CounterAtomic<N> + Send + Sync + 'static,
+{
+    fn encode_metric_value(&self) -> Vec<MetricFamily> {
+        let exemplars = self.exemplars.lock();
+        let exemplar = match &*exemplars {
+            ExemplarStorage::Empty => None,
+            ExemplarStorage::Single(exemplar) => exemplar.clone(),
+            ExemplarStorage::PerBucket(_) => unreachable!("counter uses single exemplar storage"),
+        };
+        let value = self.inner.get().into_f64();
+        drop(exemplars);
+
+        let mut families = encode_counter(value);
+        let counter = families[0].metric[0]
+            .counter
+            .as_mut()
+            .expect("counter encoder always creates a counter value");
+        counter.exemplar = finish_exemplar(exemplar.as_ref().map(StoredExemplar::encode));
+        families
     }
 }
 
--- a/foundations-metrics/src/metrics/counter.rs
+++ b/foundations-metrics/src/metrics/counter.rs
@@ -169,25 +169,9 @@ macro_rules! impl_counter_with_exemplar {
         where
             A: CounterAtomic<$value>,
         {
-            /// Returns the current exemplar.
-            pub fn exemplar(&self) -> Option<Exemplar<S, $value>> {
-                match &*self.exemplars.lock() {
-                    ExemplarStorage::Empty | ExemplarStorage::Single(None) => None,
-                    ExemplarStorage::Single(Some(StoredExemplar::$variant(exemplar))) => {
-                        Some(exemplar.clone())
-                    }
-                    ExemplarStorage::Single(Some(_)) => {
-                        unreachable!("counter uses matching exemplar value storage")
-                    }
-                    ExemplarStorage::PerBucket(_) => {
-                        unreachable!("counter uses single exemplar storage")
-                    }
-                }
-            }
-
             /// Increments the counter by `value`, records the exemplar, and
             /// returns the previous total.
-            pub fn with_exemplar(&self, label_set: S, value: $value) -> $value {
+            pub fn inc_by_with_exemplar(&self, label_set: S, value: $value) -> $value {
                 let exemplar = StoredExemplar::$variant(Exemplar::new(label_set, value, None));
                 let mut exemplars = self.exemplars.lock();
                 let previous = self.inner.inc_by(value);
--- a/foundations-metrics/src/metrics/exemplar.rs
+++ b/foundations-metrics/src/metrics/exemplar.rs
@@ -1,0 +1,745 @@
+use std::collections::HashMap;
+use std::marker::PhantomData;
+use std::sync::Arc;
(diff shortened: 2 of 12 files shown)


foundations-metrics/src/metrics/native_histogram.rs
```diff
@@ -56,6 +56,16 @@ impl NativeHistogram {
     pub fn observe(&self, value: f64) {
         self.inner.observe(value);
     }
+
+    pub(super) fn try_encode_metric_value(&self) -> Result<Vec<MetricFamily>, std::fmt::Error> {
+        // Upstream keeps native bucket state private. A cloned histogram shares
```
fisherdarling: Upstream as in `prometheus_client`? If so, let's write `prometheus_client` instead.

foundations-metrics/src/metrics/native_histogram.rs
```diff
@@ -325,7 +330,7 @@ mod tests {
 
     #[test]
     fn builder_applies_configuration() {
-        let histogram = NativeHistogramBuilder::new(1.5)
+        let histogram: NativeHistogram = NativeHistogramBuilder::new(1.5)
```
fisherdarling, line 404: Is this leftover? I'm surprised the compiler can't elide the type.

foundations-metrics/src/metrics/native_histogram.rs
ethanolchik: ok yeah i'll do that

foundations-metrics/src/metrics/native_histogram.rs
ethanolchik: `NativeHistogramBuilder` implements `MetricConstructor` for `NativeHistogram` and also `NativeHistogramWithExemplars`, and since `new_metric` doesn't have any arguments that identify the type, the compiler will error

foundations-metrics/src/metrics/exemplar.rs
```diff
@@ -0,0 +1,710 @@
...
+    marker: PhantomData<N>,
+}
+
+#[derive(Debug)]
+struct CounterWithExemplarState<S, N, A> {
+    value: A,
+    exemplar: Option<Exemplar<S, N>>,
+}
+
+impl<S, N, A> CounterWithExemplar<S, N, A>
```
fisherdarling: Think we should put these `WithExamplar` metrics in their relevant files? Like the Counter in `counter.rs`?

foundations-metrics/src/metrics/exemplar.rs
```diff
@@ -0,0 +1,710 @@
...
+///
+/// Calling [`inc_by`](Self::inc_by) with a label set replaces the previous
+/// exemplar. Calling it without a label set clears the previous exemplar. The
+/// exemplar value is the increment. [`get`](Self::get) returns the cumulative
+/// counter value and read-only access to the current exemplar. Clones share both
+/// values.
+///
+/// Exemplar labels are serialized with [`serde::Serialize`] during collection.
+#[derive(Debug)]
+pub struct CounterWithExemplar<S, N = u64, A = AtomicU64> {
```
fisherdarling: I wonder if it's possible to make this more of a smart pointer, which provides access to the inner value after providing an exemplar.

Like `WithExemplar<Counter>`

And then an api like:

```rust
impl WithExemplar<T> {
  fn with_exemplar(&self, exemplar: ...) -> &T {
     self.exemplar.lock()....
     &self.inner
  } 
  
  fn get() -> &T
}
```

Though I'm not sure if it's feasible. E.g. esp the histogram one seems difficult.

foundations-metrics/src/metrics/exemplar.rs
ethanolchik: yeah sounds good to me

foundations-metrics/src/metrics/exemplar.rs
ethanolchik: yeah I like this idea. I pushed a version using `WithExemplar<T>` with the metric-specific implementations alongside each metric type.

```rust
let counter: WithExemplar<Counter, TraceLabels> =
    WithExemplar::default();
counter.with_exemplar(labels, 1);
counter.inc();

let histogram =
    WithExemplar<Histogram, TraceLabels>::new(Histogram::new([0.5, 1.0]);
histogram.with_exemplar(labels, 0.75);
histogram.observe(0.25);
```

It also works with existing histogram builders:
```rust
let histogram: WithExemplar<Histogram, TraceLabels> =
    HistogramBuilder {
        buckets: &[0.5, 1.0],
    }
    .new_metric();

let native: WithExemplar<NativeHistogram, TraceLabels> =
    HistogramBuilder::new(1.1)
        .with_max_buckets(160)
        .new_metrics();
```

The classic histogram uses private per-bucket exemplar storage, and `with_exemplar` performs the observation so it can update the correct bucket.

have a look at the changes and let me know if the API looks good to you.

foundations-metrics/src/metrics/counter.rs
```diff
@@ -161,6 +163,81 @@ where
...
+                    }
+                    ExemplarStorage::PerBucket(_) => {
+                        unreachable!("counter uses single exemplar storage")
+                    }
+                }
+            }
+
+            /// Increments the counter by `value`, records the exemplar, and
+            /// returns the previous total.
+            pub fn with_exemplar(&self, label_set: S, value: $value) -> $value {
```
fisherdarling: Can we call this `inc_by_with_exemplar`?
Or think that's overkill?

foundations-metrics/src/metrics/exemplar.rs
```diff
@@ -0,0 +1,731 @@
...
+        let families = counter.encode_metric_value();
+        let encoded = families[0].metric[0].counter.as_ref().unwrap();
+        let exemplar = encoded.exemplar.as_ref().unwrap();
+        assert_eq!(encoded.value, Some(5.0));
+        assert_eq!(exemplar.value, Some(3.0));
+        assert_eq!(trace_id(exemplar), Some("latest"));
+        assert!(exemplar.timestamp.is_none());
+
+        // No guard is handed out, so reads and updates freely interleave. Under
+        // the previous `get() -> (N, MappedRwLockReadGuard<'_, _>)` API these
```
fisherdarling: nit: extra comment

foundations-metrics/src/metrics/histogram.rs
```diff
@@ -134,6 +143,65 @@ fn encode_snapshot(snapshot: HistogramSnapshot) -> Vec<MetricFamily> {
     }]
 }
 
+impl<S> WithExemplar<Histogram, S> {
+    /// Records `value` and stores the exemplar in the matching bucket.
+    pub fn with_exemplar(&self, label_set: S, value: f64) {
```
fisherdarling: nit: can we call this `observe_with_exemplar`

foundations-metrics/src/metrics/native_histogram.rs
```diff
@@ -56,6 +60,37 @@ impl NativeHistogram {
...
+        registry.register("native_histogram", "", self.inner.clone());
+
+        prometheus_protobuf::encode(&registry)
+            .map(|families| families.into_iter().map(convert_native_family).collect())
+    }
+}
+
+impl<S> WithExemplar<NativeHistogram, S> {
+    /// Records `value` and retains it as the latest exemplar.
+    pub fn with_exemplar(&self, label_set: S, value: f64) {
```
fisherdarling: nit: same naming here `observe_with_exemplar`

foundations-metrics/src/metrics/exemplar.rs
```diff
@@ -0,0 +1,731 @@
...
+use foundations_metrics_registry::proto;
+use parking_lot::Mutex;
+use prost_types::Timestamp;
+use serde::Serialize;
+
+use super::IntoF64;
+use crate::labels::to_label_pairs;
+use crate::validation::EXEMPLAR_SERIALIZATION_ERROR_LABEL;
+
+/// Labels and a sampled value associated with a metric observation.
```
fisherdarling, line 15: Can we add an example here, like including the traceid of a slow req with a latency histogram bucket.

foundations-metrics/src/metrics/exemplar.rs
```diff
@@ -0,0 +1,731 @@
...
+///
+/// Their fields remain private; exemplars are created by recording labeled
```
fisherdarling: I think this is an extra comment? Not sure if we need it.

foundations-metrics/src/metrics/exemplar.rs
fisherdarling: Also another Q, does this need to be public?

foundations-metrics/src/metrics/exemplar.rs
```diff
@@ -0,0 +1,731 @@
...
+pub(super) fn finish_exemplar(
+    exemplar: Option<Result<proto::Exemplar, String>>,
+) -> Option<proto::Exemplar> {
+    match exemplar {
+        Some(Ok(exemplar)) => Some(exemplar),
+        // Defer reporting to validation, after any enclosing Family lock has
+        // been released.
+        Some(Err(error)) => Some(proto::Exemplar {
+            label: vec![proto::LabelPair {
+                name: Some(EXEMPLAR_SERIALIZATION_ERROR_LABEL.to_owned()),
```
fisherdarling, line 66: This is a pretty neat trick! I wish we had results here haha, but 🤷‍♀️

foundations-metrics/src/metrics/counter.rs
ethanolchik: no its definitely not overkill, I think it makes things clearer that it performs a counter operation not just exemplar config

foundations-metrics/src/metrics/exemplar.rs
ethanolchik: no it doesn't, you're right. this was left over from testing, i'll remove it!

foundations-metrics/src/metrics/exemplar.rs
ethanolchik: Thanks! I didn't make this a `Result` because exemplar serialisation is a partial failure, and I still want to emit the metrics without the exemplars regardless. If we make it a result, we don't get this partial output option anymore. However, we can return a type like `EncodeOutput` which would be defined like:

```rust
pub struct EcnodeOutput {
    families: Vec<MetricFamily>,
    diagnostics: Vec<EncodeError>,
}
```
This would require changing `EncodeMetricValue` and its implementations, having `Family` aggregate diagnostics until its lock is released, reporting them through `NamedMetric` and removing the sentinel handling from validation.

If you think that interface is better, I can have a look at making those changes?

fisherdarling (approved):

foundations-metrics/src/metrics/exemplar.rs
ethanolchik: As discussed in DMs, this is follow-up work before foundations v6 release: replace the exemplar sentinel with structured encoding diagnostics.

Roughly:
```rust
pub struct EncodeOutput {
    pub families: Vec<MetricFamily>,
    pub diagnostics: Vec<EncodeDiagnostic>,
}

pub trait EncodeMetricValue {
    fn encode_metric_value(&self) -> Vec<MetricFamily>;

    fn encode_metric_with_diagnostics(&self) -> EncodeOutput {
        EncodeOutput {
            families: self.encode(),
            diagnostics: Vec::new(),
        }
    }
}
```
`EncodeMetric` would get an equivalent default method. `Family`, exemplar metrics, and `NamedMetric` would propagate diagnostics through the new path, while existing implementations remain unchanged.

PR #240, opened 2026-07-28
feat(foundations): use foundations-metrics as the macro backend + expose a foundations-metrics facade

ethanolchik:
## Overview

- Enables the `foundations-metrics-backend` feature by default.
- Replaces the legacy `prometheus-client`/`prometools` backend while keeping typical call sites source-compatible.
- Adds `Accept`-header negotiation to `/metrics`.
- Adds a public API for custom metrics.
- Makes `foundations-metrics` and `foundations-metrics-registry` publishable.

## Behaviour Changes

- **Exposition output:** Integral values render as `1.0`, empty `HELP` lines are omitted, and families with no serialisable rows are dropped.
- **Global registry:** Metrics registered only through `prometheus::register` are no longer collected. Register them through the foundations metrics API instead.
- **Content negotiation:** `/metrics` can return OpenMetrics text or length-delimited protobuf based on `Accept`.
- **Negotiation fallback:** Unsupported or invalid preferences fall back to text with a warning rather than failing the scrape.
- **Extra producers:** Registering any text-only `ExtraProducer` disables protobuf output process-wide.
- **Startup validation:** Invalid service label names now fail telemetry initialisation instead of producing empty scrapes.

## Public API

- Custom metrics can implement `EncodeMetricValue` and use `NamedMetric`, or implement `EncodeMetric` directly.
- Metrics are registered with `register(Box<dyn EncodeMetric>, RegistrationMetadata)`.
- The backend re-exports native histograms, `WithExemplar`, encoding traits, registration types, protobuf types, and label helpers through `telemetry::metrics`.
- Existing metric facade names now resolve to the new backend types when the feature is enabled.

## Deprecations

- Deprecates `add_extra_producer` and `ExtraProducer` in favor of `EncodeMetric` and `register`.
- The deprecated APIs remain functional with both backends but only support text output.

## Other Changes

- Applies the service name during collection, avoiding early collection permanently setting an `undefined_` prefix.
- Routes collection diagnostics through telemetry instead of stderr.
- Updates `#[metrics]` to reference metric types through the public facade.
- Adds a migration guide and documents metric name, label, and histogram validation behaviour.
- Adds the metadata required to publish both metrics crates.
- Builds the seccomp probe in a scratch directory so artifacts do not remain in `OUT_DIR`.

## Testing

- Adds coverage for custom metrics, info metrics, content negotiation, protobuf output, fallback behaviour, and startup validation.
- Run with `cargo nextest run`.
- Two tests in `foundations/tests/metrics.rs` fail under plain `cargo test` because they share the global registry; this also occurs at the base commit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,6 +85,7 @@ percent-encoding = "2.3.2"
 quote = "1.0.45"
 regex = "1.12.3"
 reqwest = { version = "0.13.2", default-features = false }
+ryu = "1.0.23"
 socket2 = { version = "0.6.3", features = ["all"] }
 syn = "2.0.117"
 serde = "1.0.228"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,8 @@ anyhow = "1.0.102"
 backtrace = "0.3.76"
 foundations = { version = "5.8.1", path = "./foundations", default-features = false }
 foundations-macros = { version = "=5.8.1", path = "./foundations-macros", default-features = false }
-foundations-metrics-registry = { version = "0.1.0", path = "./foundations-metrics-registry" }
+foundations-metrics-registry = { version = "0.1.0", path = "./foundations-metrics-registry", default-features = false }
+foundations-metrics = { version = "0.1.0", path = "./foundations-metrics", default-features = false }
 bindgen = { version = "0.72.1", default-features = false }
 cc = "1.2.61"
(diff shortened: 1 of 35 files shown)


ethanolchik:
Superseded by the focused four-part series:

- #245: info metrics & custom metrics
- #247: backend migration
- #248: exposition negotiation
- #249: publishing documentation and example

Closing this combined PR in favor of that series.

PR #241, opened 2026-07-31
ZTC-3215: Add pause syscall to seccomp SERVICE_BASICS allowlist

OmegaJak:
`std::process::exit()` has a [safeguard](https://github.com/rust-lang/rust/blob/fc0f51f5ca44d4586b0e309181382608cd6a8441/library/std/src/sys/exit.rs#L53) on linux against `libc::exit` thread unsafety which will pause a thread if it tried to exit simultaneously with another thread. This pause uses the [pause](https://man7.org/linux/man-pages/man2/pause.2.html) syscall.

If this race condition is triggered with the default seccomp configuration, the process will be killed by seccomp. This has been observed in a production environment.

This PR adds the `pause` syscall to the default seccomp allowlist to avoid this issue.

I don't think this poses any meaningful additional security risk by allowing potential malicious code to invoke it - at most, it can suspend a thread indefinitely, which I believe could already be accomplished with the `futex` syscall, which is already in the `RUST_BASICS` allowlist.

### DIFF
--- a/foundations/src/security/common_syscall_allow_lists.rs
+++ b/foundations/src/security/common_syscall_allow_lists.rs
@@ -33,6 +33,7 @@ allow_list! {
         ..RUST_BASICS,
         exit,
         exit_group,
+        pause, // pause can be invoked by std::process::exit() if multiple threads call it simultaneously
         kill if [ ArgCmp::Equal { arg_idx: 0, value: std::process::id().into() } ],
         tgkill if [ ArgCmp::Equal { arg_idx: 0, value: std::process::id().into() } ],
         getpid,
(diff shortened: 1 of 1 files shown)


nmldiegues (approved):

fisherdarling (approved):

OmegaJak:
It seems glibc pause is implemented using ppoll instead on aarch64: https://codebrowser.dev/glibc/glibc/sysdeps/unix/sysv/linux/pause.c.html#31
https://codebrowser.dev/glibc/glibc/sysdeps/unix/sysv/linux/aarch64/arch-syscall.h.html

So I've added the relevant #[cfg] to allow ppoll with those specific arguments on aarch64

TheJokr (approved):

PR #244, opened 2026-08-05
feat(metrics): Add NativeTimeHistogram, new_classic_and_native method, and third-party linux process metrics from prometheus crate

ethanolchik:
## Overview

Adds a `NativeTimeHistogram` with the same interface as `TimeHistogram` but emitting native histograms. `HistogramTimer` is generalized over the sealed `ObserveNanos` trait, so the two built-in time histogram types share the same timer implementation without exposing an extension point.

Also adds `NativeHistogram::new_classic_and_native` and an optional `classic_buckets` field on `NativeHistogramBuilder`, letting one metric carry both representations: each observation updates both, so text exposition can serve classic buckets while protobuf serves native ones.

`#[metrics]` validates literal `classic_buckets` at compile time, rejecting invalid bucket arrays the way it already does for classic histograms.

## Public API

- `NativeTimeHistogram`: `new`, `observe(nanos: u64)`, and `start_timer`.
- `HistogramTimer<H = TimeHistogram>`: the defaulted parameter keeps existing uses source-compatible.
- `ObserveNanos`: sealed and implemented only for `TimeHistogram` and `NativeTimeHistogram`.
- `NativeHistogram::new_classic_and_native`, `NativeHistogramBuilder::with_classic_buckets`, and the public `classic_buckets` field.

All are re-exported through `foundations::telemetry::metrics`.

## Linux process metrics

Preserves the Linux process metrics that the legacy backend exposed by adapting `prometheus::process_collector::ProcessCollector` to the new structured registry:

- `process_cpu_seconds_total`
- `process_resident_memory_bytes`
- `process_virtual_memory_bytes`
- `process_open_fds`
- `process_max_fds`
- `process_start_time_seconds`
- `process_threads`

The adapter converts the collector's counter and gauge families into Foundations' canonical protobuf model and registers it once per process. These metrics opt out of the service prefix and service label, preserving their existing names for dashboards and alerts, and are available in both text and protobuf exposition.

This does not restore collection from the rest of the `prometheus` global registry. Other metrics registered only there still need to migrate to the Foundations registry.

Tests cover text and protobuf output as well as service-name suppression. The full test suite, strict clippy, and Linux-target checks pass.

## Note

Merge this after #249.

### DIFF
--- a/foundations/tests/metrics_negotiation.rs
+++ b/foundations/tests/metrics_negotiation.rs
@@ -10,6 +10,8 @@
 use std::future::IntoFuture;
 use std::net::{Ipv4Addr, SocketAddr};
 
+#[cfg(target_os = "linux")]
+use foundations::telemetry::settings::{MetricsSettings, ServiceNameFormat};
 use foundations::telemetry::settings::{TelemetryServerSettings, TelemetrySettings};
 use foundations::telemetry::{TelemetryConfig, TelemetryContext};
 
@@ -21,6 +23,9 @@ const PROTOBUF_ACCEPT: &str = "application/vnd.google.protobuf;\
 
 const TEXT_ACCEPT: &str = "application/openmetrics-text;version=1.0.0;q=0.5,\
                            text/plain;version=0.0.4;q=0.4,*/*;q=0.1";
+
+#[cfg(target_os = "linux")]
+const PROCESS_CPU: &str = "process_cpu_seconds_total";
 
 #[tokio::test]
 async fn metrics_endpoint_serves_the_negotiated_format() {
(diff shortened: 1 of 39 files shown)


foundations/tests/metrics_negotiation.rs
```diff
@@ -84,6 +89,12 @@ async fn metrics_endpoint_serves_the_negotiated_format() {
         !body.ends_with(b"# EOF\n"),
         "protobuf response carries the text terminator"
     );
+    #[cfg(target_os = "linux")]
+    assert!(
+        body.windows(PROCESS_CPU.len())
```
fisherdarling, line 94: I forgot that there's no `.contains(<squence>)` on the slice api :/
https://github.com/rust-lang/libs-team/issues/499
https://github.com/rust-lang/rust/issues/134149

fisherdarling (approved):

PR #245, opened 2026-08-06
feat(metrics): add info and custom metrics

ethanolchik:
## Summary

- Add `InfoMetric` for registering one-hot build and runtime metadata as labels.
- Add collection support and registration metadata for unlabeled info metrics.
- Make the registry encoding interface public for user-defined metric implementations.
- Add `EncodeMetricValue` to public interface so custom values can use `NamedMetric` and `Family` without owning naming or family construction.
- Add integration coverage for direct encoders, value encoders, families, metadata, and registration.

## Testing

- `foundations-metrics` tests passed.
- Focused custom metric tests passed.
- Doctests passed.
- Strict clippy passed.

### DIFF
--- a/foundations-metrics-registry/src/metadata.rs
+++ b/foundations-metrics-registry/src/metadata.rs
@@ -11,6 +11,15 @@ pub struct RegistrationMetadata {
 
     /// Whether to suppress the service-name prefix for this metric.
     pub unprefixed: bool,
+
+    /// Whether to suppress the service-name label for this metric.
+    ///
+    /// Independent of [`unprefixed`](Self::unprefixed): a metric that opts out
+    /// of the service-name prefix is still labelled with the service name under
+    /// [`ServiceNameFormat::LabelWithName`]. Info metrics opt out of both.
+    ///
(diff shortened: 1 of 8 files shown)


foundations-metrics-registry/src/metadata.rs
```diff
@@ -11,6 +11,15 @@ pub struct RegistrationMetadata {
 
     /// Whether to suppress the service-name prefix for this metric.
     pub unprefixed: bool,
+
+    /// Whether to suppress the service-name label for this metric.
```
fisherdarling, line 22: small nit: can we add an example here in the text format of what this enabled? Like basically like a `before` -> `after`.

fisherdarling: Also in `unprefixed` too. Just for a visual repr of what changes.

fisherdarling (approved):

foundations-metrics-registry/src/metadata.rs
ethanolchik: yeah sure, just done that now.

PR #247, opened 2026-08-06
feat(foundations): migrate the metrics facade to foundations-metrics

ethanolchik:
## Summary

- Enable `foundations-metrics-backend` by default and route the public metrics facade through `foundations-metrics`.
- Update `#[metrics]` expansion to use the public facade while retaining the legacy backend feature path.
- Re-export the new backend types, including info metrics, custom metric APIs, native histograms, and `WithExemplar`.
- Update telemetry and hook tests for the backend-neutral facade.

## Testing

- Full nextest suite passed.
- Strict clippy passed.
- Legacy no-default-features backend check passed.

### DIFF
--- a/foundations-metrics/src/encoding/text.rs
+++ b/foundations-metrics/src/encoding/text.rs
@@ -13,9 +13,9 @@ pub const OPENMETRICS_CONTENT_TYPE: &str =
 
 /// Encodes metric families as UTF-8 OpenMetrics text.
 ///
-/// Label names are always quoted. Metric names outside the legacy Prometheus
-/// grammar use the quoted metric-name form. Serve the output with
-/// [`OPENMETRICS_CONTENT_TYPE`] so scrapers retain UTF-8 names.
+/// Metric and label names outside their legacy Prometheus grammars use quoted
+/// forms. Serve the output with [`OPENMETRICS_CONTENT_TYPE`] so scrapers retain
+/// UTF-8 names.
 pub fn encode_to_text(families: &[MetricFamily]) -> String {
     let mut output = String::new();
 
@@ -94,7 +94,7 @@ fn encode_metric(output: &mut String, name: &str, metric_type: MetricType, metri
                 "",
                 metric,
                 None,
-                SampleValue::Float(counter.value.unwrap_or_default()),
+                SampleValue::LegacyCompatibleFloat(counter.value.unwrap_or_default()),
                 counter.exemplar.as_ref(),
             );
         }
@@ -108,7 +108,7 @@ fn encode_metric(output: &mut String, name: &str, metric_type: MetricType, metri
                 name,
                 "",
                 metric,
-                SampleValue::Float(gauge.value.unwrap_or_default()),
+                SampleValue::LegacyCompatibleFloat(gauge.value.unwrap_or_default()),
             );
         }
         MetricType::Summary => {
@@ -266,6 +266,9 @@ fn has_native_buckets(histogram: &Histogram) -> bool {
 #[derive(Clone, Copy)]
 enum SampleValue {
     Float(f64),
+    // Removes `.0` from integral counters & gauges which protobuf produces
+    // to preserve compatibility with older foundations versions.
+    LegacyCompatibleFloat(f64),
     Unsigned(u64),
 }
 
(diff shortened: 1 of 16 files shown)


foundations/src/telemetry/metrics/internal.rs
```diff
@@ -218,6 +206,26 @@ pub fn wrap_metric(metric: impl SendSyncEncodeMetric + 'static) -> Box<dyn SendS
     Box::new(RewindErrorEncode(metric))
 }
 
+/// Registers a metric declared through the [`metrics`](super::metrics) macro.
+///
+/// `full_name` is unused here: the exported name is composed by the registries
+/// from the service prefix, the `subsystem` sub-registry, and `name`.
+pub fn register_metric<M>(
```
fisherdarling, line 213: We might be able to combine the two functions with a `if cfg!(...) {}` block. I think I like that, cause then both backends are visually basically a toggle between two different ways to register things.

foundations/src/telemetry/metrics/mod.rs
```diff
@@ -76,16 +162,16 @@ fn report_nonfatal_collect_error(err: &dyn Display) {
...
+/// Supported metric types are reexported from this module for convenience:
 ///
 /// * [`Counter`]
-/// * [`CounterWithExemplar`]
 /// * [`Gauge`]
 /// * [`Histogram`]
-/// * [`HistogramWithExemplars`]
 /// * [`TimeHistogram`]
 ///
+/// To attach exemplars, wrap any of the above in [`WithExemplar<T, S>`], which
```
fisherdarling, line 172: docs nit: can you add a small example (doesn't need to compile)? Just a hint.

foundations/src/telemetry/metrics/mod.rs
```diff
@@ -412,12 +507,23 @@ impl MetricConstructor<Histogram> for HistogramBuilder {
     }
 }
 
+#[cfg(not(feature = "foundations-metrics-backend"))]
 impl<S> MetricConstructor<HistogramWithExemplars<S>> for HistogramBuilder {
```
fisherdarling, line 516: Do we need to a type alias for `HistogramWithExemplar`? Would that break things?

Ah, breaking things is ok and we're allowing people to opt-in to breakage with a feature.

foundations/tests/metrics.rs
```diff
@@ -3,6 +3,11 @@ use foundations::telemetry::metrics::{self, Counter, Family, metrics};
 use foundations::telemetry::settings::{MetricsSettings, ServiceNameFormat, TelemetrySettings};
 use foundations::telemetry::{TelemetryConfig, TelemetryContext};
 
+#[cfg(not(feature = "foundations-metrics-backend"))]
+const ONE: &str = "1";
+#[cfg(feature = "foundations-metrics-backend")]
```
fisherdarling: I presume this is an expected breaking change, right?
I know many repos assert on the specific strings, is there a way to unify this between backends? Think it's worth it?

foundations/src/telemetry/metrics/internal.rs
ethanolchik: I tried this, but `if cfg!()` still typechecks both branches. That fails because `foundations_metrics` is optional and the two backends require different generic bounds. I could move them into a new `registration.rs` file and re-export this in `internal_backend.rs` and `internal.rs`, or just keep it the way it is at the moment.

foundations/src/telemetry/metrics/mod.rs
ethanolchik: yeah pretty much. `HistogramWithExemplar` is only available in the legacy backend but once people migrate over I don't see a reason to swap back to the legacy implementation

foundations/tests/metrics.rs
ethanolchik: Hmm I didn't know this. The new backend stores values as protobuf doubles so `1` becomes `1.0`. OpenMetrics considers them equivalent, but if people are asserting on the strings themselves, I think its best to render it without the `.0` for compatibility.

foundations/src/telemetry/metrics/internal.rs
fisherdarling: That's fine, we can leave it as is here. We'll remove the old code hopefully 🔜

foundations-metrics/src/encoding/text.rs
```diff
@@ -266,6 +266,9 @@ fn has_native_buckets(histogram: &Histogram) -> bool {
 #[derive(Clone, Copy)]
 enum SampleValue {
     Float(f64),
```
fisherdarling, line 268: Is this unused now?

fisherdarling (approved):

foundations-metrics/src/encoding/text.rs
ethanolchik: It's still used by histograms & untyped metrics

PR #248, opened 2026-08-06
feat(foundations): negotiate metrics scrape formats

ethanolchik:
## Summary

- Negotiate OpenMetrics text and length-delimited protobuf responses from the metrics endpoint using `Accept` preferences.
- Fall back to text for unsupported or invalid preferences, and when legacy text-only extra producers are registered.
- Apply the configured service name during collection and reject service label names that cannot be encoded.
- Add integration coverage for negotiation, extra-producer fallback, protobuf output, and startup validation.

## Testing

- Full nextest suite passed.
- Strict clippy passed.
- Legacy no-default-features backend check passed.

### DIFF
--- a/foundations/src/telemetry/metrics/mod.rs
+++ b/foundations/src/telemetry/metrics/mod.rs
@@ -75,6 +75,28 @@ mod backend {
 
 pub use backend::*;
 
+/// Translates telemetry settings into collection options.
+///
+/// The service name is only known once telemetry is initialized, so it is
+/// applied at collection time rather than at registration time.
+#[cfg(feature = "foundations-metrics-backend")]
+fn collection_options(settings: &MetricsSettings) -> foundations_metrics::CollectionOptions<'_> {
+    let service_name_format = match &settings.service_name_format {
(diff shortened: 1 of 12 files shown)


foundations/src/telemetry/metrics/mod.rs
```diff
@@ -129,6 +137,205 @@ pub fn collect(settings: &MetricsSettings) -> Result<String> {
...
+        if format == ScrapeFormat::Protobuf {
+            let families = foundations_metrics::collect(collection_options(settings));
+
+            return Ok((
+                format.content_type(),
+                foundations_metrics::encode_to_protobuf(&families),
+            ));
+        }
+
+        Ok((format.content_type(), collect(settings)?.into_bytes()))
```
fisherdarling, line 308: collect herre is the "legacy" `collect` right? Which is always text format?

ethanolchik: yeah exactly, its the text-only `collect`. but internally it uses the new `foundations_metrics::collect` function and then just uses the text encoder.

fisherdarling (approved):

PR #252, opened 2026-08-11
Add an owned UserSpan handle for user tracing

mar-cf:
The user-tracing API is built around an ambient scope stack: `span()` parents
under whatever is current and hands back an RAII guard that pops on drop. The
scope stack is a `ThreadLocal` and its guard is deliberately `!Send`, so a span
can never be held across an await point.

That model can't express a per-request span in a service whose request pipeline
is a state machine of separate callbacks rather than a call tree. The span
outlives every individual callback, so there is no lexical scope for it to live
in, and the only `Send` escape hatch available today is `into_context()` — which
is meant for propagating into a future, and drags a snapshot of the whole
ambient `TelemetryContext` along with it.

`UserSpan` is an owned, `Send` handle to a user span, detached from the scope
stack. It can live in a struct field, cross threads, and be finished in a
different callback from the one that started it. The span is reported when the
handle drops.

### DIFF
--- a/foundations/src/telemetry/tracing/mod.rs
+++ b/foundations/src/telemetry/tracing/mod.rs
@@ -25,9 +25,11 @@ mod output_otlp_uds;
 mod traceparent;
 
 use self::init::TracingHarness;
-#[cfg(feature = "user-tracing")]
-use self::internal::current_user_span;
 use self::internal::{SharedSpan, create_span, current_span, shared_span, span_trace_id};
+#[cfg(feature = "user-tracing")]
+use self::internal::{
+    SharedSpanHandle, child_user_span, current_user_span, start_user_trace, user_shared_span,
+};
 use super::TelemetryContext;
 use super::scope::Scope;
 use std::borrow::Cow;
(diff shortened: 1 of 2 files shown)


foundations/src/telemetry/tracing/mod.rs
```diff
@@ -265,6 +267,125 @@ impl UserSpanScope {
...
+                state.span_id(),
+                state.flags()
+            ))
+        })
+    }
+
+    /// Makes this span current for the lifetime of the returned scope, so the ambient helpers in
+    /// [`user_tracing`] target it.
+    ///
+    /// The returned scope is `!Send` and must not be held across an `await` point.
```
TheJokr, line 384: I see that UserSpan does not impl Clone, I assume that is deliberate to control its lifetime?

In which case, it might be worth noting here that this extends the lifetime of the span until the scope is dropped.

TheJokr (approved):

foundations/src/telemetry/tracing/mod.rs
mar-cf: Correct.

Updated to fix the incorrect claim.

PR #253, opened 2026-08-12
refactor(metrics): finalize public API for publishing

ethanolchik:
## Overview
- Keeps the `foundations-metrics-backend` feature flag off by default
- Separates negotiation from collection, deprecates the legacy `collect`
- Remove `ObserveNanos` from public API
- Re-exports backend builders and sets the correct crate versions.

### DIFF
--- a/foundations-metrics/Cargo.toml
+++ b/foundations-metrics/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "foundations-metrics"
 description = "Metric types and encoders for the Foundations stack."
-version = "0.1.0"
+version = "0.1.0-beta1"
 readme = "README.md"
 repository = { workspace = true }
 edition = { workspace = true }
--- a/foundations-metrics/Cargo.toml
+++ b/foundations-metrics/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "foundations-metrics"
 description = "Metric types and encoders for the Foundations stack."
-version = "0.1.0-beta1"
+version = "0.1.0-beta.1"
 readme = "README.md"
 repository = { workspace = true }
 edition = { workspace = true }
--- a/foundations/src/telemetry/metrics/mod.rs
+++ b/foundations/src/telemetry/metrics/mod.rs
@@ -4,7 +4,14 @@
 //! - Use [`metrics`] macro to define regular metrics.
 //! - Use [`report_info`] function to register service information metrics (metrics, whose value is
 //!   persistent during the service lifetime, e.g. software version).
-//! - Use [`collect_format`] method to obtain metrics report programmatically.
+#![cfg_attr(
+    feature = "foundations-metrics-backend",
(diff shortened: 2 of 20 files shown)


foundations/src/telemetry/metrics/mod.rs
```diff
@@ -112,7 +115,53 @@ fn collect_registered_metrics(
...
+/// did not encode.
+///
+/// # Errors
+///
+/// Fails when `format` cannot represent everything this process exposes.
+/// [`ScrapeFormat::Protobuf`] requires the `foundations-metrics-backend`
+/// feature, and cannot carry the output of a registered extra producer, which is
+/// opaque text; [`allow_protobuf`] reports whether it may be asked for.
+/// [`ScrapeFormat::fallback`] never fails for this reason.
+pub fn collect_format(format: ScrapeFormat, settings: &MetricsSettings) -> Result<Vec<u8>> {
```
TheJokr, line 147: i'd prefer if we made this only available with the new backend, to avoid people shooting themselves in the foot by manually asking for protobuf encoding.

then we make the `collect` deprecation conditional as well (`#[cfg_attr(feature = "...", deprecated = "...")]`) to avoid giving warnings where there is no alternative

foundations/src/telemetry/metrics/mod.rs
```diff
@@ -149,203 +198,30 @@ pub fn collect(settings: &MetricsSettings) -> Result<String> {
...
+/// `false` while any extra producer is registered: those hand over opaque text,
+/// which cannot be transcoded into the protobuf data model, so serving protobuf
+/// would silently drop every series they emit.
 ///
-/// Evaluated per scrape because producers may be registered at any point.
+/// Evaluate it per scrape, because producers may be registered at any point.
 #[cfg(feature = "foundations-metrics-backend")]
 #[allow(deprecated)]
-fn protobuf_available() -> bool {
+pub fn allow_protobuf() -> bool {
```
TheJokr, line 237: with the collect_format change, this can be new-backend-only as well. and lets mark it deprecated from the get-go, since it will be removed at the same time as ExtraProducer.

foundations/src/telemetry/metrics/mod.rs
TheJokr: please make the ExtraProducer deprecations conditional on the new backend as well. i want to avoid giving people build warnings that they cannot fix with the old backend.

foundations-metrics/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "foundations-metrics"
 description = "Metric types and encoders for the Foundations stack."
-version = "0.1.0"
+version = "0.1.0-beta1"
```
TheJokr: after checking semver.org now, apparently this should be `beta.1`. sorry for giving you the wrong string originally!

```suggestion
version = "0.1.0-beta.1"
```

TheJokr (approved):

PR #257, opened 2026-08-20
feat(telemetry): USDT probes for tracing spans end

jsitnicki:
## Problem

Code bases which use `foundations` are already instrumented with tracing spans, but there is no easy way to measure the duration of a specific span with a BPF-based tracer, like `bpftrace` or `ebpf_exporter`. One would have to craft custom uprobes at code offsets that shift with every build, and match span start/end events across await points and thread hops.

## Solution

`span_with_probe!("span::name")` — a drop-in replacement for `tracing::span("...")` that additionally emits a per-span USDT probe at span end, `span_end_<sanitized name>`, receiving the span duration (ns) as an argument.

Key features:
- Attaching is zero-cost for everything else: each call site gets a probe semaphore that tracers bump on attach; unattached spans don't even record a timestamp.
- Works regardless of span sampling — probes fire with tracing disabled.
- linux/x86_64 only; other platforms degrade to plain `tracing::span`. aarch64 support to follow.
- Comes with a demo workload + bpftrace script, `examples/span_with_probe`, to show the intended use.

### DIFF
--- a/foundations-macros/src/span_with_probe.rs
+++ b/foundations-macros/src/span_with_probe.rs
@@ -1,0 +1,319 @@
+use crate::common::parse_optional_trailing_meta_list;
+use darling::FromMeta;
+use proc_macro::TokenStream;
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
+use syn::parse::{Parse, ParseStream};
+use syn::{LitStr, Path, parse_quote};
+
+struct Args {
+    span_name: LitStr,
+    options: Options,
+}
+
+#[derive(FromMeta)]
+pub(crate) struct Options {
+    #[darling(default = "Options::default_crate_path")]
+    crate_path: Path,
+
+    #[darling(default = "Options::default_usdt_provider")]
+    usdt_provider: LitStr,
+}
+
+impl Options {
+    fn default_crate_path() -> Path {
+        parse_quote!(::foundations)
+    }
+
+    pub(crate) fn default_usdt_provider() -> LitStr {
+        let provider =
+            std::env::var(USDT_PROVIDER_ENV_VAR).unwrap_or_else(|_| "foundations".into());
+
+        LitStr::new(&provider, proc_macro2::Span::call_site())
+    }
+}
+
+/// Environment variable that overrides the default USDT provider at compile
+/// time; settable per project via `[env]` in `.cargo/config.toml`.
+pub(crate) const USDT_PROVIDER_ENV_VAR: &str = "FOUNDATIONS_USDT_PROVIDER";
+
+/// Stable-Rust substitute for the unstable `proc_macro::tracked_env`: the
+/// `option_env!` read lands in the consumer crate's dep-info, so Cargo
+/// rebuilds the call site (rerunning the macro) when the override changes.
+pub(crate) fn track_provider_env() -> TokenStream2 {
+    let env_var = USDT_PROVIDER_ENV_VAR;
+
+    quote!(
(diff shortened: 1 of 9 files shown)


foundations-macros/src/span_with_probe.rs
```diff
@@ -0,0 +1,219 @@
...
+            }
+        })
+        .collect()
+}
+
+/// `stapsdt` note + NOP, adapted from probe-rs' `sdt!` (x86_64, SystemTap
+/// semaphore in `.probes`). The two `{}` operands are the semaphore symbol
+/// and the duration argument.
+///
+/// `usdt_provider` must not contain `:` (bpftrace attach points are
```
TheJokr: please add a check for this when expanding the macro

jsitnicki:
Per our discussion off-line, reworked the changes to support `span_fn`.

This called for a completely different approach than in `span_with_probe!` case.
Wrapping an existing `span_fn` macro with another macro (`span_fn_with_probe`?)
seemed silly. I've added a new option instead - `end_probe`. `span_fn` shares
what's possible with `span_with_probe!`.

Other changes:
- We can't have `:` chars in USDT provider or probe name because libbpf's
  auto-attach routine won't accept them. See
  https://github.com/libbpf/libbpf/blob/f7081a6baf3f54949aacb8c2fc11bb30783b83e9/src/libbpf.c#L13418
  To keep things simple just replace any non-alphanumeric chars with `_`.
- Let users set USDT provider per project with `FOUNDATIONS_USDT_PROVIDER` env var.

Open question: Do we want to mark this as unstable and gate it be a crate
feature for now? How flexible are we on public API changes?

What's next? As a follow-up I will work on `probe-rs` to:

1. Propose the public API change that would let us drop the `asm!` snippets in
   `foundations`.
2. Add `aarch64` support.
3. Add optimized syscall-based USDT (`nop,nop10` combo) probe support.

foundations/src/telemetry/tracing/mod.rs
```diff
@@ -235,6 +263,20 @@ impl SpanScope {
...
+
+    /// Arms the span's USDT probe when `record_probe_start` is true (the
+    /// span's probe semaphore, a `static` in the `.probes` section bumped by
+    /// the tracer on attach, is non-zero). `end_probe` is the address of a
+    /// per-span function containing the probe's NOP placeholder; it is called
+    /// with the span duration in nanoseconds when the last clone of the span
+    /// drops. Arming records the start timestamp unconditionally with respect
+    /// to sampling, so probes work even when span tracing is disabled.
+    #[doc(hidden)]
+    pub fn __arm_probe(&mut self, record_probe_start: bool, end_probe: fn(u64)) {
```
TheJokr: please make this `#[inline(always)]` to avoid the function call in the common (disabled) case. same for `DualSpanScope::__arm_probe`.

TheJokr:
LGTM other than that 1 nit!

foundations/src/telemetry/tracing/mod.rs
jsitnicki: I should also just pull up the `record_probe_start` check into `__arm_probe` caller.

jsitnicki:
* pull up check if probe is enabled out of `__arm_probe`
* annotate `__arm_probe` with `#[inline(always)]`
* update sample output in `examples/README.md`

jsitnicki:
* Wrap `SEC("usdt/<path>:<provider>:<name>")` in backticks to make `rustdoc` happy

fisherdarling (approved):

fisherdarling (approved):

PR #259, opened 2026-08-26
Release 5.9.2

TheJokr:
### Fixed
- #258: With `foundations-metrics-backend` enabled, the `process` metrics implementation relied on an accessor in the `prometheus` crate that is not present when the `protobuf` feature is enabled. This change enables the `protobuf` feature in `prometheus` instead and uses a different accessor.
- #260: `NativeHistogram`, when scraped in protobuf format, now correctly exposes an upper bound of `f64::INFINITY` rather than `f64::MAX`.

### DIFF
--- a/foundations-metrics/RELEASE_NOTES.md
+++ b/foundations-metrics/RELEASE_NOTES.md
@@ -1,5 +1,6 @@
 
-0.1.0-beta.2
+0.1.0-beta.3
+- 2026-08-26 fix(metrics): terminate classic histogram buckets with +Inf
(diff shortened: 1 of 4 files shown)


foundations-metrics/RELEASE_NOTES.md
```diff
@@ -1,5 +1,6 @@
 
-0.1.0-beta.2
+0.1.0-beta.3
```
TheJokr, line 2: i put the tag for `beta.2` on a commit that didn't touch anything in `foundations-metrics/`, so git-cliff doesnt pick it up here

dotjs (approved):

PR #263, opened 2026-09-03
Cache `SharedSpan::is_sampled` for deferred user tracing spans

TheJokr:
Relaxed atomic load and stores are just regular load and stores on x86 and arm64, so this adds zero cost for regular spans and allows us to avoid some more lock accesses for deferred user tracing spans.


fisherdarling (approved):

fisherdarling (changes requested):
didn't mean to stamp! whoops

fisherdarling (approved):

PR #268, opened 2026-09-15
feat(telemetry): configure service name for user traces

wperron:
Use UserTracingSettings::service_name for OTLP resources when user tracing is enabled, omit the internal service version, and reject enabled exporters with an empty name.

When user tracing is disabled, initialization still exits before exporter setup and the existing internal tracing path continues to use ServiceInfo name and version unchanged.


Bongajum (approved):
Nothing to previous @bongajum

foundations/src/telemetry/settings/user_tracing.rs
```diff
@@ -11,6 +11,9 @@ pub struct UserTracingSettings {
     #[serde(default = "UserTracingSettings::default_enabled")]
     pub enabled: bool,
 
+    /// Publicly visible service name attached to user traces.
+    pub service_name: String,
```
TheJokr, line 16: please add to the docstring that this must be non-empty if user tracing is enabled

TheJokr (approved):

TheJokr (approved):

PR #270, opened 2026-09-18
feat(telemetry): allow extra arguments with span end probes

jsitnicki:
Sometimes one span covers more than one kind of operation. Let users pass    extra arguments to the span-end probe so tracers can split its latency by  category.


oga35767-eng (approved):

oga35767-eng:
👍

